PR #5, opened 2015-02-13
Some ascii tests

protocol7:


src/test/java/com/spotify/folsom/client/ascii/DeleteRequestTest.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2014-2015 Spotify AB
```
spkrka: Not sure, but maybe this should be only 2015?

spkrka:
:+1:

src/test/java/com/spotify/folsom/client/ascii/DeleteRequestTest.java
protocol7: I'll fix

PR #6, opened 2015-02-14
Add coveralls

spkrka:


spkrka:
@protocol7

spkrka:
Slight annoyance - the coverage runs once for each JDK, which seems unnecessary. Not sure how to configure travis to only run it once.

PR #9, opened 2015-02-14
Add test for large values and make RoundRobinMemcacheClient distribute requests more evenly

spkrka:


protocol7:
+1

src/main/java/com/spotify/folsom/roundrobin/RoundRobinMemcacheClient.java
```diff
@@ -30,8 +31,6 @@
  * host/port which might be a reasonable thing to do if the IO operations is a bottleneck.
  */
 public class RoundRobinMemcacheClient extends AbstractMultiMemcacheClient {
-  private static final int MAX_COUNTER = Integer.MAX_VALUE / 2;
-
   private final AtomicInteger counter = new AtomicInteger(0);
   private final List<RawMemcacheClient> clients;
   private final int numClients;
```
rschildmeijer, line 36: Not related to this PR.

Is the explicit clients.size() value (stored in numClient) an intentional optimization?

spkrka, line 36: Yes, I suppose so, but I am not sure it actually optimizes anything.

PR #11, opened 2015-02-15
Test with even larger values, move code to right place

spkrka:


spkrka:
@protocol7 Maybe I put too many changes in the same PR now... but it would be nice to have some review on the API change - it might leak too many abstractions i guess, but i couldn't think of a cleaner way to expose a more complete connection state

protocol7:
+1, merge away!

PR #12, opened 2015-02-16
Add tests for handling misbehaving servers

spkrka:


spkrka:
@danielnorberg and @protocol7  - I changed some handling for closing client connections so that would be nice to have reviewed.

spkrka:
@protocol7 This sort of _is_ the test for AsciiMemcacheDecoder. It is at least _a_ test, but we should probably have some more. :)

PR #17, opened 2015-02-16
Simplify MemcacheClientBuilder.newStringClient() by defaulting to UTF-8

nresare:
Other charsets can still be used by explicitly providing a
StringTranscoder to he public constructor

I believe this change will simplify usage for new users while still retaining the flexibility (and possibly a slight speed optimisation) for power users.

### DIFF
--- a/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
+++ b/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
@@ -16,6 +16,7 @@
 
 package com.spotify.folsom;
 
+import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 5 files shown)


spkrka:
No strong opinion, I will wait for @protocol7 to chime in.

src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -82,12 +82,15 @@
   }
 
   /**
-   * Create a client builder with a basic string transcoder.
-   * @param charset the charser to use for encoding/decoding strings
+   * Create a client builder with a basic string transcoder using the UTF-8
+   * Charset. To build a client with a different charset, provide a StringTranscoder
+   * constructed with a different charset in the public MemcacheClientBuilder
```
protocol7: Having to explain this indicates a problem with the API. How about adding an overloaded method without the charset (I agree on promoting UTF-8 as the "default" charset) while keeping the existing one?

nresare: I guess it's a matter of opinion how much you want to discourage custom charsets. I don't think that the documentation hint I added necessarily indicates a problem, but take your point.

nresare:
I guess code comments on outdated versions gets lost in the noise. Anyway, patch updated along the lines of @protocol7 suggestion.

PR #25, opened 2015-02-18
Add option of choosing key charset, enable keys with non-ascii characters

spkrka:


nresare:
I looked around a bit for a test case that actually tests non-ascii keys but failed to find it. Perhaps we should have that? (Or perhaps I'm blind)

spkrka:
Yes, that should be added before we merge this. I wanted to get some more general feedback on code architecture/design first.

nresare:
If you just want a review of the overall thinking, I have looked through the code and it seems reasonable to me. The fact that the number of parameters is pretty high in many places doesn't matter that much in my opinion as they are at the lower levels of the code.

josefalcon:
:+1:

src/main/java/com/spotify/folsom/client/Request.java
```diff
@@ -15,30 +15,37 @@
...
     this.opaque = (opaque << 8) & 0xFFFFFF00;
-    Utils.validateKey(key);
   }
 
   protected static ByteBuf toBuffer(final ByteBufAllocator alloc, ByteBuffer dst) {
     return toBuffer(alloc, dst, 0);
   }
 
-  public String getKey() {
+  public byte[] getKey() {
```
protocol7, line 48: Not in scope for this PR, but APIs that expose byte[] is a problem. Should we use/do something like ByteString?

src/main/java/com/spotify/folsom/client/Request.java
```diff
@@ -65,4 +72,32 @@ protected static ByteBuf toBuffer(final ByteBufAllocator alloc, ByteBuffer dst,
   public int getOpaque() {
     return opaque;
   }
+
+  @VisibleForTesting
+  static byte[] encodeKey(String key, Charset charset) {
+    checkNotNull(key, "key");
```
protocol7, line 78: Check charset as well

src/main/java/com/spotify/folsom/client/Request.java
spkrka, line 48: True, though this class is not really directly exposed to regular users of the library.

We could of course try to improve it by making it package private and move all the classes that need it into the same package. Which might make sense anyway.

src/main/java/com/spotify/folsom/client/Request.java
spkrka, line 78: Sounds like a good idea (though it can never be null if it's used through the MemcacheClientBuilder class)

src/main/java/com/spotify/folsom/client/Request.java
```diff
@@ -65,4 +72,32 @@ protected static ByteBuf toBuffer(final ByteBufAllocator alloc, ByteBuffer dst,
...
+    if (length > 250) {
+      throw new IllegalArgumentException("Key is too long: " + key);
+    }
+    if (length <= 0) {
+      throw new IllegalArgumentException("Key is empty");
+    }
+    for (int i = 0; i < length; i++) {
+      final byte c = keyBytes[i];
+      if (c >= 0 && c <= 32) {
+        throw new IllegalArgumentException("Invalid key: " + key);
```
protocol7, line 91: Hmm, this is tricky. With multi-byte characters, bytes can end up in that range, e.g. ASCII characters in UTF-16. I wonder how/if that works with memcached

src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
```diff
@@ -43,24 +43,26 @@
...
     this.rawMemcacheClient = rawMemcacheClient;
     this.metrics = metrics;
     this.valueTranscoder = valueTranscoder;
+    this.charset = charset;
     this.transformerUtil = new TransformerUtil<>(valueTranscoder);
   }
 
   @Override
   public ListenableFuture<MemcacheStatus> set(final String key, final V value, final int ttl) {
-    final String keyBytes = encodeKey(key);
```
protocol7, line 59: checkNotNull(key)?

src/main/java/com/spotify/folsom/client/binary/NoopRequest.java
```diff
@@ -27,12 +28,21 @@
 public class NoopRequest extends BinaryRequest<Void> {
 
   public NoopRequest(final int opaque) {
-    super("", opaque);
+    super("X", Charsets.US_ASCII, opaque);
```
protocol7, line 32: hmm, at least a comment here :)

src/main/java/com/spotify/folsom/client/Request.java
protocol7, line 48: Agreed

spkrka, line 48: Agree to the second statement I presume? :)

src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
spkrka, line 59: Not needed, the request class will check that.

src/main/java/com/spotify/folsom/client/binary/NoopRequest.java
spkrka, line 32: Will add that!
Something like:
// The key has to be valid to pass the Request contract, but it's not actually used later on.

src/main/java/com/spotify/folsom/client/Request.java
protocol7, line 91: A quick test indicated that memcached errs in that case, so they really seem to treat each byte as a character. So, using UTF-16 or similar would probably work _really_ poorly and the above check is the correct one.

spkrka, line 91: Good question! I am fairly sure memcached treats the input as a stream of 8-bit characters, and only explicitly disallowing space and some control values.

This probably implies that UTF-16 won't work for memcached, assuming some valid utf-16 character is represented as 0x??20 (where ?? can be anything)

spkrka, line 91: Or even more common I guess - any character represented as 0x00?? (is there where the regular ASCII fits in UTF-16?)

spkrka, line 91: It does indeed seem like memcached just looks for 0x20 (' ') to tokenize things, so a key may not contain that character. It also uses '\0' and '\r' and '\n' so those are probably also a bad idea to use. We could relax our current key byte requirement from [33, 255] to everything except 0, \r, \n and ' ', but I am not sure how much that actually improves things. We still won't be able to claim that we support encodings such as UTF-16

https://github.com/memcached/memcached/blob/master/memcached.c#L2443

spkrka, line 91: Should we add a whitelist of allowed charsets? Or just add javadoc saying that some encodings may result in failed requests (at least we will quickly fail in a nice way)

spkrka:
I fixed everything except the public getKey().
Making that method (and some of the others there) not public causes a chain reaction of package changes, I think that should be a separate pull request.

nresare:
One idea I got reading your discussion is that perhaps it is best to just always use UTF-8. UTF-8 is backwards compatible with ASCII, doesn't encode any characters with null bytes or overlapping ascii control characters. That way the code could probably be simplified while supporting the full range of java Strings.

spkrka:
Perhaps there is some value in support these charsets though? http://en.wikipedia.org/wiki/ISO/IEC_8859

In any case, I don't think the code complexity is that much worse now, in fact I think some parts are slightly better (key validation and ascii decoder)

PR #26, opened 2015-02-18
support for flush commmand

srirampmalladi:
Change-Id: I381a7b537784a8d2b864caa7067d1ff579571cef

### DIFF
--- a/src/main/java/com/spotify/folsom/client/ascii/FlushAllRequest.java
+++ b/src/main/java/com/spotify/folsom/client/ascii/FlushAllRequest.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2015 Spotify AB
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.spotify.folsom.client.ascii;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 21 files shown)


src/main/java/com/spotify/folsom/client/binary/FlushAllRequest.java
```diff
@@ -0,0 +1,55 @@
...
+  private final int delay;
+
+  public FlushAllRequest(final int delay, final int opaque) {
+    super("", opaque);
+    this.delay = delay;
+  }
+
+  @Override
+  public ByteBuf writeRequest(final ByteBufAllocator alloc, final ByteBuffer dst) {
+    writeBinaryHeader(dst, OpCode.FLUSH, 0, 4, 4, 0, opaque);
```
spkrka, line 39: the extraLength should be either 0 or 4 here I guess, depending on the value of delay

spkrka:
Nice! I would like to see some added IntegrationTest method too, and possibly an extra test for MisbehavingServerTest (should hopefully be very simple to add)

src/test/java/com/spotify/folsom/client/binary/FlushAllRequestTest.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) 2014-2015 Spotify AB
```
spkrka: this should only say 2015

src/main/java/com/spotify/folsom/client/ascii/FlushAllRequest.java
```diff
@@ -0,0 +1,51 @@
...
+    return toBuffer(alloc, dst);
+  }
+
+  @Override
+  protected void handle(AsciiResponse response) throws IOException {
+    AsciiResponse.Type type = response.type;
+    if (type == AsciiResponse.Type.OK) {
+      succeed(null);
+    } else {
+      throw new IOException("Unexpected line: " + type);
```
spkrka, line 50: this line could be tested with MisbehavingServerTest

src/main/java/com/spotify/folsom/client/binary/FlushAllRequest.java
srirampmalladi, line 39: Actually, I will remove the if condition in the next line so we write the delay whatever its value is.

spkrka, line 39: Sounds good!

Issue #28, opened 2015-02-19
MemcacheClientBuilderTest tests fails to connect to embedded memcached


nresare:
Running the tests in MemcacheClientBuilderTest stalls trying to reconnect to a server that doesn't answer with exponential backoff on my OSX machine. I would assume that the tests work on travis-ci due to the system memcached.

nresare:
Ah, dynamic ports! PR coming up.

PR #29, opened 2015-02-19
Make sure tests connect to embedded memcached on the proper port

nresare:
Fix to https://github.com/spotify/folsom/issues/28


spkrka:
Good catch!

spkrka:
:+1:

PR #30, opened 2015-02-19
Simplify opaque management in binary protocol

spkrka:


spkrka:
This should probably be abandoned in favor of #33

PR #32, opened 2015-02-20
Make the connect notification functionality generic

nresare:
Introduce a new super interface ObservableClient extended by both RawMemcacheClient
and the higher level interfaces.

### DIFF
--- a/src/test/java/com/spotify/folsom/KetamaIntegrationTest.java
+++ b/src/test/java/com/spotify/folsom/KetamaIntegrationTest.java
@@ -33,6 +33,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import static java.util.Arrays.asList;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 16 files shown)


src/test/java/com/spotify/folsom/KetamaIntegrationTest.java
```diff
@@ -96,7 +114,8 @@ public void setUp() throws Exception {
     } else {
       client = builder.connectBinary();
     }
-    IntegrationTest.awaitConnected(client);
+    allClientsConnected(client);
+    ConnectFuture.connectFuture(client).get();
```
spkrka: isn't this line redundant?

src/test/java/com/spotify/folsom/KetamaIntegrationTest.java
```diff
@@ -70,6 +72,22 @@ public static void tearDownClass() throws Exception {
     binaryServers.stop();
   }
 
+  public static void allClientsConnected(final MemcacheClient<?> client)
+      throws Exception{
+    final CountDownLatch done = new CountDownLatch(1);
+
+    client.registerForConnectionChanges(new ConnectionChangeListener() {
```
spkrka, line 79: maybe this should use a future similar to ConnectFuture instead of a countdownlatch?

src/test/java/com/spotify/folsom/KetamaIntegrationTest.java
nresare: nice catch. Managed to leave in the broken code as well.

src/test/java/com/spotify/folsom/KetamaIntegrationTest.java
nresare, line 79: Well, I opted for the simpler solution here. In the future I imagine that we will provide utility functions to have a future return when say 90% of a cluster is up, but I think that we should put some more thinking into the semantics of that.

With countdown latch it is easy to have it timeout as well (even though I managed to misremember the fact that it doesn't throw an exception when timing out but instead just returns false. I'll fix that in an upcoming patch)

spkrka:
:+1: good to go!

nresare:
Hmm.. git rebase happily moves a diff if it's target file has moved, which causes import errors. I didn't expect that one.

PR #33, opened 2015-02-20
Opaque generation

spkrka:


spkrka:
Ping @danielnorberg

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -250,7 +253,12 @@ public void run() {
     public void write(final ChannelHandlerContext ctx, final Object msg,
                       final ChannelPromise promise)
         throws Exception {
-      outstanding.add((Request<?>) msg);
+      Request<?> request = (Request<?>) msg;
+      if (request instanceof BinaryRequest) {
+        ((BinaryRequest) request).setOpaque(requestSequenceId.incrementAndGet());
```
spkrka: I guess this one doesn't have to be an atomicInteger - is a regular integer fine, or should it be volatile?

(AFAIK, this should all run from the same thread)

danielnorberg: A regular integer will do fine. The connection handler is single threaded.

danielnorberg:
:+1:

protocol7:
+1

PR #35, opened 2015-02-20
Partition multirequests for large requests

josefalcon:
@spkrka

### DIFF
--- a/src/test/java/com/spotify/folsom/IntegrationTest.java
+++ b/src/test/java/com/spotify/folsom/IntegrationTest.java
@@ -16,11 +16,13 @@
 
 package com.spotify.folsom;
 
+import com.google.common.base.Function;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.common.net.HostAndPort;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.spotify.folsom.client.NoopMetrics;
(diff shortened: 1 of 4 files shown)


src/test/java/com/spotify/folsom/IntegrationTest.java
```diff
@@ -310,13 +310,22 @@ public void testGetKeyNotExist() throws Throwable {
...
+      final String key = "key-" + i;
+      keys.add(key);
+      client.set(key, key, TTL);
+    }
+
+    final ListenableFuture<List<String>> resultsFuture = client.get(keys);
+    final List<String> results = resultsFuture.get();
+    assertEquals(keys, results);
+
+    for (String key : keys) {
```
spkrka: Do the deletes in a finally-block in case the test fails.

src/test/java/com/spotify/folsom/IntegrationTest.java
```diff
@@ -310,13 +310,22 @@ public void testGetKeyNotExist() throws Throwable {
...
-  public void testTooLargeMultiget() throws Exception {
+  @Test
+  public void testPartitionMultiget() throws Exception {
     List<String> keys = Lists.newArrayList();
-    for (int i = 0; i < 1000; i++) {
-      keys.add("key-" + i);
+    for (int i = 0; i < 500; i++) {
+      final String key = "key-" + i;
+      keys.add(key);
+      client.set(key, key, TTL);
```
spkrka: Need to wait for the set to finish if we want the test to be deterministic (but we can at least wait for all of them at the same time if you store them in a list of futures.

spkrka:
Looks good, we can merge as soon as my inlined comments are addressed I think.

josefalcon:
@spkrka Added the waits

spkrka:
@danielnorberg - There are some interesting errors in this log: https://travis-ci.org/spotify/folsom/jobs/51637155 - not sure if there's something wrong in folsom itself or in jmemcached, the embedded server we use.

Probably caused by this:
15:17:48.024 [defaultRawMemcacheClient-1-26] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. Enable advanced leak reporting to find out where the leak occurred. To enable advanced leak reporting, specify the JVM option '-Dio.netty.leakDetectionLevel=advanced' or call ResourceLeakDetector.setLevel()

PR #42, opened 2015-02-27
Add optimization for cancelled requests

spkrka:


spkrka:
Ping @danielnorberg and @protocol7

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -16,6 +16,7 @@
 
 package com.spotify.folsom.client;
 
+import com.google.common.annotations.VisibleForTesting;
```
danielnorberg, line 19: I think usage of `VisibleForTesting` is usually a sign of design/layering problems.

Expose a metric instead?

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -89,6 +90,16 @@
    */
   private int requestSequenceId = 0;
 
+  /**
+   * Used to count number of writes that were skipped because the request was already cancelled
+   */
+  private int skippedWrites;
```
danielnorberg: This is not thread safe, so the tests might be flaky.

spkrka: Yes, I'll change it to volatile. I wanted to avoid that for performance reasons, but I guess most requests won't be cancelled.

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
spkrka, line 19: Agreed, but a metric seems overkill. The metric is not interesting for any other use case than testing I think.

danielnorberg:
LTGM but seems some test failed.

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -259,6 +270,10 @@ public void write(final ChannelHandlerContext ctx, final Object msg,
                       final ChannelPromise promise)
         throws Exception {
       Request<?> request = (Request<?>) msg;
+      if (request.isCancelled()) {
```
danielnorberg, line 273: I guess that the pending counter would need to be decremented here.

spkrka, line 273: Yes, good point.

Though I think it's perhaps best to simply abandon this change, I don't think the added complexity can be justified.

danielnorberg, line 273: Yeah, it'd probably be rare that requests would cancelled before they're written anyway.

PR #44, opened 2015-03-10
add metrics gauge for pending request count

mattnworb:
Allows for monitoring of the number of pending operations.

Changes to support this are more extensive than might be expected since
unlike other metrics that can be "pushed" into the Metrics
implementation, there needs to be a way for the Metrics implementation
to "pull" the pending operation count when it is interested in reporting
the values (otherwise the DefaultRawMemcacheClient would have to inform
the Metrics implementation every single time the pending operation
counter changed value).

Also needed to make changes to pass the Metrics instance down to
the DefaultRawMemcacheClient layer since the pending count is only
kept there.

### DIFF
--- a/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
+++ b/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
@@ -26,6 +26,7 @@ import com.spotify.folsom.ConnectFuture;
 import com.spotify.folsom.Metrics;
 import com.spotify.folsom.RawMemcacheClient;
 import com.spotify.folsom.client.DefaultRawMemcacheClient;
+import com.spotify.folsom.client.NoopMetrics;
 import com.spotify.folsom.client.NotConnectedClient;
 import com.spotify.folsom.client.Request;
 import org.slf4j.Logger;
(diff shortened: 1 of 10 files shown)


src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
```diff
@@ -62,13 +63,14 @@ public ReconnectingClient(final BackoffFunction backoffFunction,
                             final boolean binary,
                             final Executor executor,
                             final long timeoutMillis,
-                            final Charset charset) {
+                            final Charset charset,
+                            final Metrics metrics) {
```
spkrka, line 80: this is a breaking change... but then again, we're still at 0.x

src/test/java/com/spotify/folsom/client/SlowStaticServer.java
```diff
@@ -0,0 +1,83 @@
+package com.spotify.folsom.client;
```
spkrka, line 17: copyright header?

spkrka:
Looks ok to me, but maybe @protocol7 has some good feedback too since he loves metrics!

src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
mattnworb, line 80: If there is concern about anyone using the ReconnectingClient constructor directly instead of a MemcacheBuilder method, I could simply add an overload for the constructor so that the constructor with the old signature passes a NoOpMetrics

spkrka, line 80: Yes, I was going to suggest that, but then I realized that we don't have a strong need to maintain such strong API guarantees anyway at this time, so perhaps it's better to keep it clean. (But we should probably bump the minor version instead of patch version)

mattnworb:
addressed both of @spkrka's comments in 71d1313 and 9d65540

spkrka:
Looks good to me, but I want someone else to have opinions first.

mattnworb:
@spkrka makes sense, was just mentioning addressing those for any future readers.

I rebased this against master since it had conflicts with checkstyle changes to DefaultRawMemcacheClientTest.

mattnworb:
bump can anyone else review this?

PR #45, opened 2015-03-19
Run checkstyle against test sources

mattnworb:
This would have helped me catch forgetting to add a copyright header to a new test source file created in #44.

Most changes in the code are around whitespace and lines longer than 100 characters. IMO having to figure out how to wrap too-long lines when the cutoff is relatively short like 100 chars gets into all sorts of bikeshedding about style for wrapping lines, but I figured as long as the project has checkstyle rules defined they might as well be applied to tests as well.


spkrka:
Very nice! And fixing all the existing tests to comply is very much appreciated.

Issue #47, opened 2015-04-02
IntegrationTest#testPartitionMultiget is flaky


danielnorberg:
Seems like jmemcached might be flaky. Most test runs complete successfully but when it fails, the root cause seems to be an `IndexOutOfBoundsException` in jmemcached.

```
14:37:24.635 [New I/O server worker #1-1] ERROR c.t.j.p.t.MemcachedResponseEncoder - error
java.lang.IndexOutOfBoundsException: null
    at org.jboss.netty.buffer.SlicedChannelBuffer.checkIndex(SlicedChannelBuffer.java:222) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.buffer.SlicedChannelBuffer.getByte(SlicedChannelBuffer.java:93) ~[netty-3.2.3.Final.jar:na]
    at com.thimbleware.jmemcached.protocol.text.MemcachedCommandDecoder.eol(MemcachedCommandDecoder.java:53) ~[jmemcached-core-1.0.0.jar:na]
    at com.thimbleware.jmemcached.protocol.text.MemcachedCommandDecoder.decode(MemcachedCommandDecoder.java:68) ~[jmemcached-core-1.0.0.jar:na]
    at org.jboss.netty.handler.codec.frame.FrameDecoder.callDecode(FrameDecoder.java:282) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.handler.codec.frame.FrameDecoder.messageReceived(FrameDecoder.java:214) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.channel.Channels.fireMessageReceived(Channels.java:274) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.channel.Channels.fireMessageReceived(Channels.java:261) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.channel.socket.nio.NioWorker.read(NioWorker.java:350) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.channel.socket.nio.NioWorker.processSelectedKeys(NioWorker.java:281) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.channel.socket.nio.NioWorker.run(NioWorker.java:201) ~[netty-3.2.3.Final.jar:na]
    at org.jboss.netty.util.internal.IoWorkerRunnable.run(IoWorkerRunnable.java:46) [netty-3.2.3.Final.jar:na]
[...]

PR #48, opened 2015-04-02
patch jmemcached to fix ascii fragmentation issue

danielnorberg:
Our patched MemcachedCommandDecoder is on the classpath before the jmemcached jar so it takes precedence. Who said java can't do monkey patching?

https://code.google.com/p/jmemcache-daemon/issues/detail?id=32

https://github.com/spotify/folsom/issues/47


spkrka:
Sort of a nasty fix, but if it works I approve. Hopefully we can kill it off if we upgrade to a new version of jmemcached

danielnorberg:
jmemcached seems unmaintained so it's unlikely we'll see a new version of jmemcached.

danielnorberg:
We're on latest. And it was released four years ago.

PR #50, opened 2015-04-10
Use dns-java 3.0 which adds support for dns ttl

spkrka:
Also remove the option of setting dns refresh period
manually.

### DIFF
--- a/src/main/java/com/spotify/folsom/ketama/SrvKetamaClient.java
+++ b/src/main/java/com/spotify/folsom/ketama/SrvKetamaClient.java
@@ -22,6 +22,7 @@ import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.spotify.dns.DnsSrvResolver;
+import com.spotify.dns.LookupResult;
 import com.spotify.folsom.AbstractRawMemcacheClient;
 import com.spotify.folsom.ConnectFuture;
 import com.spotify.folsom.ConnectionChangeListener;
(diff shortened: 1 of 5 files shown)


protocol7:
+1

src/main/java/com/spotify/folsom/ketama/SrvKetamaClient.java
```diff
@@ -99,17 +98,33 @@ public void updateDNS() {
...
+                .sortedCopy(hosts);
+        if (!newAddresses.equals(addresses)) {
+          addresses = newAddresses;
+          log.info("Connecting to " + newAddresses);
+          List<AddressAndClient> addressAndClients = getAddressesAndClients(newAddresses);
+          setPendingClient(addressAndClients);
+        }
+      } finally {
+        long delay = clamp(MIN_DNS_WAIT_TIME, MAX_DNS_WAIT_TIME, ttl);
+        refreshJob = this.executor.schedule(refreshRunnable, delay, TimeUnit.SECONDS);
```
danielnorberg, line 124: Seems we'd wait 1 hour before attempting to resolve again if there was no records found. That'll happen from time to time in prod due to infra breakage.

How about a special case where we retry earlier when no records were found?

spkrka, line 124: True, didn't think about that case. Maybe we should keep the old setting then, but have it be treated as either the max value or the default if there are no hosts found.

spkrka, line 124: fixed now with a force push.

danielnorberg:
:+1:

PR #55, opened 2015-10-07
Add consistent support for 0 TTL

josefalcon:
According to the [spec](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L163), a TTL of 0 should be treated as 'do not expire'.

@spkrka

### DIFF
--- a/src/test/java/com/spotify/folsom/client/binary/SetRequestTest.java
+++ b/src/test/java/com/spotify/folsom/client/binary/SetRequestTest.java
@@ -20,12 +20,12 @@ import com.google.common.base.Charsets;
 import com.google.common.collect.Lists;
 import com.spotify.folsom.client.MemcacheEncoder;
 import com.spotify.folsom.client.OpCode;
-
+import io.netty.buffer.ByteBuf;
 import org.junit.Test;
 
 import java.util.List;
 
-import io.netty.buffer.ByteBuf;
+import static org.junit.Assert.assertEquals;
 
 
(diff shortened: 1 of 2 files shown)


spkrka:
I'll try to take a look at this soon!

mattnworb:
looks like CircleCI failed because of a random problem cloning the source, I rebuild it and looks all green this time.

src/test/java/com/spotify/folsom/client/binary/SetRequestTest.java
```diff
@@ -42,13 +42,22 @@ public void testBufferCas() throws Exception {
...
+  @Test
+  public void testZeroTTL() throws Exception {
+    verifySetRequest(0, 258);
+  }
+
   private void verifySetRequest(long cas) throws Exception {
+    verifySetRequest(1000, cas);
+  }
+
+  private void verifySetRequest(int ttl, long cas) throws Exception {
```
spkrka, line 54: I would prefer consistent ordering of arguments (i.e. put cas first like the method above)

protocol7:
+1

spkrka:
I took a deeper look at the code. You're right that it should be fixed.

Currently we only do this timestamp calculation in the binary protocol, not in the ascii protocol. We should probably fix that. Furthermore, we should add a safeguard to always convert TTL's larger than 60_60_24*30 to timestamps - See http://leonid.shevtsov.me/en/memcached-expiration-time-a-horrible-api-fail

Issue #60, opened 2016-02-26
Binary client throws IllegalStateException on retries?


odsod:
Example stack trace:

```
com.spotify.folsom.MemcacheClosedException: opaque may not be set more than one
 at com.spotify.folsom.client.DefaultRawMemcacheClient.send(DefaultRawMemcacheClient.java:194)
 at com.spotify.folsom.reconnect.ReconnectingClient.send(ReconnectingClient.java:94)
 at com.spotify.folsom.ketama.KetamaMemcacheClient.sendSplitRequest(KetamaMemcacheClient.java:98)
 at com.google.common.util.concurrent.ExecutionList.executeListener(ExecutionList.java:156)
 at com.google.common.util.concurrent.ExecutionList.execute(ExecutionList.java:145)
 at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:202)
 at com.google.common.util.concurrent.Futures$ChainingListenableFuture.run(Futures.java:902)
 at com.google.common.util.concurrent.Futures$1$1.run(Futures.java:635)
 at com.google.common.util.concurrent.ExecutionList.execute(ExecutionList.java:145)
 at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:202)
 at com.google.common.util.concurrent.Futures$CombinedFuture.access$400(Futures.java:1608)
 at com.google.common.util.concurrent.Futures$CombinedFuture$2.run(Futures.java:1686)
 at com.google.common.util.concurrent.MoreExecutors$DirectExecutor.execute(MoreExecutors.java:457)
 at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
 at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
 at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
 at com.spotify.folsom.client.DefaultRawMemcacheClient.send(DefaultRawMemcacheClient.java:194)
 at com.spotify.folsom.reconnect.ReconnectingClient.send(ReconnectingClient.java:94)
 at com.spotify.folsom.ketama.KetamaMemcacheClient.sendSplitRequest(KetamaMemcacheClient.java:98)
 at com.spotify.folsom.ketama.KetamaMemcacheClient.send(KetamaMemcacheClient.java:69)
 at com.spotify.folsom.retry.RetryingClient$1.create(RetryingClient.java:51)
[...]

PR #61, opened 2016-02-27
fix binary req retries - allow setOpaque() twice

danielnorberg:
https://github.com/spotify/folsom/issues/60

The retrying client retries by re-sending the original request. The underlying binary memcache protocol implementation did not allow resending requests. 
- Remove the restriction of only allowing one call to setOpaque().
- Add tests to verify that both the raw ascii and raw binary clients allow request resends.


protocol7:
+1

Issue #67, opened 2017-03-14
DefaultRawMemcacheClient logs error but does now throw


jvaleo:
Whenever there is a timeout an error is logged but there is not exception thrown. This means that the caller can't catch anything and do something else. The use case for us is to catch a timeout exception, throw a custom timeout exception our framework provides which ticks metrics, logs, et cetera. This is what we do for all clients we use (Cassandra, RPC, et cetera). 

If you're OK with this change, we can submit a patch for this behavior.

https://github.com/spotify/folsom/blob/master/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java#L285

jvaleo:
(Ping)

danielnorberg:
@jvaleo Looks like all outstanding request futures will be failed when the channel is closed on timeout:

https://github.com/spotify/folsom/blob/master/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java#L319

Are you seeing futures not failing on timeout?

jvaleo:
@danielnorberg  So the issue is that an error is logged every time the client times out, but it's difficult to take an action on those timeouts (vs another exception) because all client exceptions cause outstanding futures to throw the same exception type (with a different message). We want to monitor or take a specific action on a timeout and its not ideal to have to catch the MemcacheClosedException and check for the string "Timeout". Ideally, the exception thrown by the future would extend TimeoutException (or the future itself could timeout) so that we could manage them cleanly.

danielnorberg:
@jvaleo Right, that sounds useful.

jvaleo:
@danielnorberg  Is that something we can just PR?

protocol7:
@jvaleo Feel free to push a PR and we'll happily review it

mniebla:
Something I was interested too...so I took a crack at it, here you go: https://github.com/spotify/folsom/pull/71

PR #71, opened 2017-05-03
Throw MemcacheTimeoutException instead of logging ERROR for request timeout

mniebla:
Reference https://github.com/spotify/folsom/issues/67

Alters `DefaultRawMemcacheClient` to throw a `MemcacheTimeoutException` instead of logging ERROR for request timeout.

Timeouts dont necessarily indicate a problem with the remote system (maybe a packet was lost).

So rather than logging this as an ERROR, fails the request with `MemcacheTimeoutException`. Upstream caller can determine their course of action with the timeout: log it, retry, fallback to origin, etc.

The new exception class `MemcacheTimeoutException` extends `MemcacheClosedException`...because the client still closes the conn...so its a special case of this broader exception.

Exception is created with the following message:

```
Request timeout after 100.247631 ms: [id: 0xf3810f83, /127.0.0.1:60562 => localhost/127.0.0.1:60560] com.spotify.folsom.client.ascii.GetRequest@2bb20830
```

Elapsed time is calculated by adding an `elapsed()` method to the `TimeoutChecker`

Finally, retained the log  message in `DefaultRawMemcacheClient`, but only as a **debug**...so if someone **really** wants to see the timeouts at this level, they can configure the logger for the class.

### DIFF
--- a/src/main/java/com/spotify/folsom/MemcacheTimeoutException.java
+++ b/src/main/java/com/spotify/folsom/MemcacheTimeoutException.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2014-2015 Spotify AB
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.spotify.folsom;
+
+public class MemcacheTimeoutException extends MemcacheClosedException {
+  public MemcacheTimeoutException(final String message) {
+    super(message);
+  }
+}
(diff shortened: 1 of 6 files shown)


danielnorberg:
Will all outstanding requests fail with MemcacheTimeoutException or only the "head"?

mniebla:
I **believe** what happens here is that `head` fails with `MemcacheTimeoutException`. Outstanding requests get handled by the `channelInactive` method...which will set all pending to `MemcacheClosedException`. 

Someone correct me if I read this wrong...

mniebla:
If the intent here is to treat requests that fail because of a timeout...differently than requests that fail because of something memcached actively failed...maybe all the pending requests should get the same treatment...

thoughts @danielnorberg ?

jvaleo:
My $.02 is treating the timeouts differently than requests that fail because of another reason makes total sense. I opened an issue specially for that as we (and others) want to be able to handle timeouts differently than other exceptions.

mniebla:
@jvaleo agree with that generally - the issue is though...

1. there are **pending** requests in the queue behind the request that timed-out
2. when we timeout the **head** request...all the **pending** requests will also fail because we close the connection.

So, how do we fail the **pending** requests? Should those also be timeouts? They did not actually timeout...but they failed because an a request in front of them timed out. So, they should fail as they did before as `MemcacheClosedException`, which is pretty descriptive of why they failed.

However, if the intent here is to treat the requests that fail due to timeouts, in a different manner...then we should fail all the **pending** requests....like we fail the **head** request....as a `MemcacheTimeoutException`, no?

PR #72, opened 2017-06-12
Shutdown the client on failure to connect and lost of connection

AlexPayment:
I think it wouldn't hurt to call `shutdown()` on the internal client in the `ReconnectingClient` when a connection failed to be establish and when the connection is lost. This will assure that the underlying channel get closed properly and potentially prevents any kind of connection leakage.

Is there any cons in doing this?

### DIFF
--- a/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
+++ b/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
@@ -129,6 +129,7 @@ public class ReconnectingClient extends AbstractRawMemcacheClient {
             @Override
             public void onSuccess(final Void ignore) {
               log.info("Lost connection to {}", address);
+              client.shutdown();
               notifyConnectionChange();
               if (stayConnected) {
                 retry();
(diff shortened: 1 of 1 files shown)


alexeyOnGitHub:
this is a critical fix. without it Folsom library leaks memory on connection failures.

this fix solved OutOfMemory error in our app under load test when Memcached was down and the app had thousands of failed connections to Memcached  per second. all the leaked memory was in byte[] arrays inside MemcachedEncoder class.

spkrka:
Thanks for the report. I think however that this is the wrong place to fix the bug in - the reconnecting client should only create a new client once the old client is already disconnected - and that should take care of cleaning up itself.

Therefore I propose a different change to make client cleanup more robust. See https://github.com/spotify/folsom/pull/76

Feel free to give feedback on that (or more feedback in here) - I may be completely misunderstanding the problem!

spkrka:
I have now release folsom 0.8.0 with some bug fixes. Can you try that and see if it fixes your issues?
You can also use the the new `Utils.getGlobalConnectionCount()` to keep track of number of active connections - if it's continuously growing there's still a problem

PR #76, opened 2017-08-21
Make sure to always close netty channel when shutting down a raw client

spkrka:


src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -74,6 +74,8 @@
   private static final EventLoopGroup EVENT_LOOP_GROUP =
       new NioEventLoopGroup(0, DAEMON_THREAD_FACTORY);
 
+  static final AtomicInteger GLOBAL_CONNECTION_COUNT = new AtomicInteger();
```
protocol7: I'd wrap this in a static getter returning an int rather than exposing the raw AtomicInteger (even with default access)

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -414,12 +418,14 @@ private void setDisconnected(Throwable cause) {
   }
 
   private void setDisconnected(String message) {
+    // assert message != null
```
protocol7: Remove

src/main/java/com/spotify/folsom/client/Utils.java
```diff
@@ -76,4 +76,11 @@ public static int ttlToExpiration(final int ttl) {
     };
   }
 
+  /**
+   * A counter of all currently connected clients. This can be useful to detect connection leaks
+   */
+  public static int getGlobalConnectionCount() {
+    return DefaultRawMemcacheClient.GLOBAL_CONNECTION_COUNT.get();
```
protocol7: How about exposing this as part of Metrics?

spkrka: I will look into it!

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
spkrka: Ok

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
spkrka: That's sane!

src/test/java/com/spotify/folsom/IntegrationTest.java
```diff
@@ -118,6 +118,8 @@ public static void tearDownClass() throws Exception {
 
   @Before
   public void setUp() throws Exception {
+    assumeTrue(0 == Utils.getGlobalConnectionCount());
```
protocol7: Should this really be an assumption (i.e ignoring the test if not true), wouldn't we rather want the test to fail in that case as it would highlight something being broken?

src/test/java/com/spotify/folsom/MemcacheClientBuilderTest.java
```diff
@@ -47,36 +47,51 @@ public void tearDown() throws Exception {
   public void testValidLatin1() throws Exception {
     AsciiMemcacheClient<String> client = MemcacheClientBuilder.newStringClient()
             .withKeyCharset(Charsets.ISO_8859_1)
-            .withAddress(HostAndPort.fromParts("localhost", server.getPort()))
+            .withAddress(HostAndPort.fromParts("127.0.0.6", server.getPort()))
```
protocol7, line 50: Why are we using 127.0.0.6, as opposed to 127.0.0.1? For TCP tuple uniqueness? Shouldn't the port already solve that? Seems a bit messy to have to ensure unique IPs in tests.

src/test/java/com/spotify/folsom/IntegrationTest.java
spkrka: True I suppose, but then something else should have failed earlier and we can fast fail the rest of the tests.

src/test/java/com/spotify/folsom/MemcacheClientBuilderTest.java
spkrka: I wanted to simplify associated log lines with test code - ports are usually dynamic

src/test/java/com/spotify/folsom/IntegrationTest.java
spkrka: But... I guess we would fastfail anyway with assert - i'll fix

PR #79, opened 2017-09-08
Workaround Guava deprecation of getHostText

spkrka:


spkrka:
Probably not a good idea to merge this. Alternatives include:
* Break the API and stop using HostAndPort completely
* Upgrade guava to 20.0 and start using the new method
* Do nothing, wait for people to complain

PR #80, opened 2017-11-15
[WIP] Move to Java 8 CompletableFutures

ssarne:
This PR change the Folsom library to use Java 8 Completable Futures.

It is currently work in progress, only created to share the initial changes.

### DIFF
--- a/src/main/java/com/spotify/folsom/AsciiMemcacheClient.java
+++ b/src/main/java/com/spotify/folsom/AsciiMemcacheClient.java
@@ -15,7 +15,7 @@
  */
 package com.spotify.folsom;
 
-import com.google.common.util.concurrent.ListenableFuture;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * A memcache client using the ascii protocol
@@ -27,18 +27,18 @@ public interface AsciiMemcacheClient<V> extends MemcacheClient<V> {
   /**
    * Increment a counter for the provided key
    *
-   * @param key     The key, must not be null
-   * @param by      The value to increment the counter by
+   * @param key The key, must not be null
+   * @param by  The value to increment the counter by
    * @return A future representing completion of the request, with the new value of the counter
    */
-  ListenableFuture<Long> incr(String key, long by);
+  CompletableFuture<Long> incr(String key, long by);
 
   /**
    * Decrement a counter for the provided key
    *
-   * @param key     The key, must not be null
-   * @param by      The value to decrement the counter by
+   * @param key The key, must not be null
+   * @param by  The value to decrement the counter by
    * @return A future representing completion of the request, with the new value of the counter
    */
-  ListenableFuture<Long> decr(String key, long by);
+  CompletableFuture<Long> decr(String key, long by);
 }
--- a/src/main/java/com/spotify/folsom/AsciiMemcacheClient.java
+++ b/src/main/java/com/spotify/folsom/AsciiMemcacheClient.java
@@ -15,7 +15,7 @@
  */
 package com.spotify.folsom;
 
-import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 
 /**
  * A memcache client using the ascii protocol
@@ -31,7 +31,7 @@ public interface AsciiMemcacheClient<V> extends MemcacheClient<V> {
    * @param by  The value to increment the counter by
    * @return A future representing completion of the request, with the new value of the counter
    */
-  CompletableFuture<Long> incr(String key, long by);
+  CompletionStage<Long> incr(String key, long by);
 
   /**
    * Decrement a counter for the provided key
@@ -40,5 +40,5 @@ public interface AsciiMemcacheClient<V> extends MemcacheClient<V> {
    * @param by  The value to decrement the counter by
    * @return A future representing completion of the request, with the new value of the counter
    */
-  CompletableFuture<Long> decr(String key, long by);
+  CompletionStage<Long> decr(String key, long by);
 }
--- a/src/main/java/com/spotify/folsom/ConnectFuture.java
+++ b/src/main/java/com/spotify/folsom/ConnectFuture.java
@@ -15,11 +15,10 @@
  */
 package com.spotify.folsom;
 
-import com.google.common.util.concurrent.AbstractFuture;
-import com.google.common.util.concurrent.ListenableFuture;
+import java.util.concurrent.CompletableFuture;
 
 public class ConnectFuture
-        extends AbstractFuture<Void>
+        extends CompletableFuture<Void>
         implements ConnectionChangeListener {
 
   private final boolean awaitedState;
@@ -35,11 +34,11 @@ public class ConnectFuture
     check(client);
   }
 
-  public static ListenableFuture<Void> disconnectFuture(ObservableClient client) {
+  public static CompletableFuture<Void> disconnectFuture(ObservableClient client) {
     return new ConnectFuture(client, false);
   }
 
-  public static ListenableFuture<Void> connectFuture(ObservableClient client) {
+  public static CompletableFuture<Void> connectFuture(ObservableClient client) {
     return new ConnectFuture(client, true);
   }
 
@@ -50,9 +49,8 @@ public class ConnectFuture
 
   private void check(ObservableClient client) {
     if (awaitedState == client.isConnected()) {
-      if (set(null)) {
-        client.unregisterForConnectionChanges(this);
-      }
+      complete(null);
+      client.unregisterForConnectionChanges(this);
     }
   }
 }
--- a/src/main/java/com/spotify/folsom/ConnectFuture.java
+++ b/src/main/java/com/spotify/folsom/ConnectFuture.java
@@ -16,7 +16,6 @@
 package com.spotify.folsom;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 
 public class ConnectFuture
     extends CompletableFuture<Void>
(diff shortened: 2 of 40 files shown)


src/main/java/com/spotify/folsom/ketama/KetamaMemcacheClient.java
```diff
@@ -89,41 +89,23 @@ private RawMemcacheClient getClient(final byte[] key) {
...
       final RawMemcacheClient client = entry.getKey();
-      ListenableFuture<List<T>> send = client.send(subRequest);
+      CompletableFuture<List<T>> send = client.send(subRequest);
       futures.put(client, send);
     }
-    final ListenableFuture<List<List<T>>> allFutures = Futures.allAsList(futures.values());
-    return Utils.transform(allFutures, new Assembler<>(futures, routing2));
-  }
-
-  private static class Assembler<T, R> implements Function<List<List<T>>, List<T>> {
```
ssarne: Did I completely miss the point here, is the assembler doing anything more than just merging the result?

protocol7:
This is great stuff, thanks for doing the work. I'd like @spkrka to give input as he knows this stuff way better than I do. A question first though, shouldn't the API expose CompletionStage as opposed to CompletableFuture?

ssarne:
Thanks, agree. It should expose CompletionStages. Will fix.

src/main/java/com/spotify/folsom/AsciiMemcacheClient.java
```diff
@@ -27,18 +27,18 @@
...
    * Increment a counter for the provided key
    *
-   * @param key     The key, must not be null
-   * @param by      The value to increment the counter by
+   * @param key The key, must not be null
+   * @param by  The value to increment the counter by
    * @return A future representing completion of the request, with the new value of the counter
    */
-  ListenableFuture<Long> incr(String key, long by);
+  CompletableFuture<Long> incr(String key, long by);
```
spkrka: as @protocol7 already said - CompletionStage for all api return values

src/main/java/com/spotify/folsom/ConnectFuture.java
```diff
@@ -35,11 +34,11 @@ private ConnectFuture(ObservableClient client, boolean awaitedState) {
     check(client);
   }
 
-  public static ListenableFuture<Void> disconnectFuture(ObservableClient client) {
+  public static CompletableFuture<Void> disconnectFuture(ObservableClient client) {
```
spkrka, line 37: Hm, here it might make sense to keep CompletableFuture instead of CompletionStage though.

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -189,43 +187,54 @@ public int getOutstandingRequests() {
...
+  public <T> CompletableFuture<T> send(final Request<T> request) {
 
+    if (!tryIncrementPending()) {
       // Do the disconnect check in here instead of outside
       // to get better performance in the happy case.
       String disconnectReason = this.disconnectReason.get();
       if (disconnectReason != null) {
-        MemcacheClosedException exception = new MemcacheClosedException(disconnectReason);
-        return onExecutor(Futures.<T>immediateFailedFuture(exception));
+        return newExceptionallyCompletedFuture(new MemcacheClosedException(disconnectReason));
```
spkrka, line 198: why remove onExecutor? Is the default for Guava futures to run callbacks on the same thread, while Java 8 always run on a common FJP as default? If so, then I am ok with this.

My main concern is users running callbacks on netty threads

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -189,43 +187,54 @@ public int getOutstandingRequests() {
...
+              "too many outstanding requests"));
     }
+
     if (request instanceof SetRequest) {
       SetRequest setRequest = (SetRequest) request;
       byte[] value = setRequest.getValue();
       if (value.length > maxSetLength) {
-        return (ListenableFuture<T>) onExecutor(
-                Futures.immediateFuture(MemcacheStatus.VALUE_TOO_LARGE));
+        // Is this cast correct. Shouldn't return type be CompletableFuture<MemcacheStatus>?
```
spkrka: Good question, I don't remember why that cast was needed. Will need to investigate.

src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
```diff
@@ -62,150 +64,156 @@ public DefaultAsciiMemcacheClient(final RawMemcacheClient rawMemcacheClient,
...
-    final ListenableFuture<List<GetResult<byte[]>>> future =
-            Utils.transform(Futures.allAsList(futureList), Utils.<GetResult<byte[]>>flatten());
+        Lists.partition(keys, MemcacheEncoder.MAX_MULTIGET_SIZE);
+
+    final List<CompletableFuture<List<GetResult<byte[]>>>> futuresList =
+        keyPartition.stream()
+            .map(part -> rawMemcacheClient.send(MultigetRequest.create(part, charset, withCas)))
+            .collect(Collectors.toList());
+
+    final CompletableFuture<List<GetResult<byte[]>>> [] futures =
```
spkrka, line 205: Use https://github.com/spotify/completable-futures#allaslist instead

src/main/java/com/spotify/folsom/ketama/KetamaMemcacheClient.java
spkrka: It merges it, but in the correct order.

src/main/java/com/spotify/folsom/ketama/SrvKetamaClient.java
```diff
@@ -191,30 +190,26 @@ private void setPendingClient(List<AddressAndClient> addressAndClients) {
...
+            }
+
+            oldClient = currentClient;
+            currentClient = pendingClient;
+            pendingClient = null;
           }
+          notifyConnectionChange();
+          executor.schedule(new ShutdownJob(oldClient), shutdownDelay, shutdownUnit);
+        })
+        .exceptionally(t -> {
```
spkrka, line 208: I guess we don't even need to include the `.exceptionally`  then

src/main/java/com/spotify/folsom/retry/RetryingClient.java
```diff
@@ -35,23 +36,39 @@
...
   @Override
-  public <T> ListenableFuture<T> send(final Request<T> request) {
-    final ListenableFuture<T> future = delegate.send(request);
-    return Futures.catchingAsync(future, MemcacheClosedException.class,
-        new AsyncFunction<MemcacheClosedException, T>() {
-          @Override
-          public ListenableFuture<T> apply(final MemcacheClosedException e) {
-            if (delegate.isConnected()) {
-              return delegate.send(request);
+  public <T> CompletableFuture<T> send(final Request<T> request) {
```
spkrka: This seems like a significant refactoring - why? Can we do it as a separate change or is it required for the java8 migration?

src/test/java/com/spotify/folsom/IntegrationTest.java
```diff
@@ -322,23 +323,24 @@ public void testPartitionMultiget() throws Exception {
       keys.add("key-" + i);
     }
 
-    final List<ListenableFuture<MemcacheStatus>> futures = Lists.newArrayList();
+    // final List<CompletableFuture<MemcacheStatus>> futures = Lists.newArrayList();
```
spkrka: Comment seems redundant

src/main/java/com/spotify/folsom/retry/RetryingClient.java
```diff
@@ -35,23 +36,39 @@
...
+    sendHelper(request, promise, 0);
+    return promise;
+  }
+
+  private <T> void sendHelper(final Request<T> request,
+                              final CompletableFuture<T> promise,
+                              final int retries) {
+    delegate.send(request)
+        .whenComplete((r, t) -> {
+          if (r != null) {
```
spkrka, line 61: will t be wrapped as a `CompletionException` here? Might be worth taking into account

src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
ssarne: Yeah, they are run on the default fork-join pool. Most likely (but not guaranteed) on the same thread, since it is a bit biased when deploying next task. 

The idea was that it is simpler without an additional pool.

danielnorberg: This depends on the method (e.g. `thenFoo` vs `thenFooAsync`) and whether the future is already completed or not.

In this case, the future is already complete so `thenFoo(cb)` will invoke the callback on the caller thread and `thenFooAsync(cb)` will [typically](http://hg.openjdk.java.net/jdk9/jdk9/jdk/file/tip/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#l429) invoke it on the common FJP. I think that's fine for all immediately completed futures (i.e. errors).

For futures that are not yet completed, `thenFoo(cb)` will cause the callback to be invoked on the _completing_ thread. So by keeping the `onExecutor` wrapping of the actual request future on [line 215](https://github.com/spotify/folsom/pull/80/files#diff-1b1a61f1bf891fde7e8125494305e141R215) we avoid user callbacks being invoked on netty IO threads.

protocol7:
Closing as this was implemented in #84

PR #86, opened 2017-12-29
add semantic-metrics implementation

mattnworb:
This is a port of internal code we developed at Spotify that implements Folsom's `Metrics` interface with [semantic-metrics][].

The module was written before semantic-metrics was open sourced, so it was not added to folsom at the time of writing but kept internal.

There isn't any real reason to keep it internal any longer so this change adds it to folsom so that things like versions or API changes can be managed by having this module in the same source tree as folsom itself.

To add this to this repo but keep its usage as optional for any folsom users, I restructured the project as a multimodule project first. I also removed the java8 profile from the folsom module (and moved the checkstyle plugins into the parent pom) as it seems unnecessary to have those plugins be optional now that the source is java8-only.

[semantic-metrics]: https://github.com/spotify/semantic-metrics


mattnworb:
I am never sure if the `nexus-staging-maven-plugin` plugin belongs in `<plugins>` or `<pluginManagement>` for a multimodule build, I left it in the parent's `<plugins>` but I'm not sure how to test it without actually doing a release.

mattnworb:
Ah, I see the `java8` profile was there to prevent the checkstyle plugin from being run when built with `oraclejdk9`. I'll add that back.

folsom-semantic-metrics/src/main/java/com/spotify/folsom/client/SemanticFolsomMetrics.java
```diff
@@ -0,0 +1,343 @@
...
+import com.spotify.folsom.Metrics;
+import com.spotify.metrics.core.MetricId;
+import com.spotify.metrics.core.SemanticMetricRegistry;
+
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * {@link com.spotify.folsom.Metrics} implementation using semantic-metrics.
+ * Mostly a port of the YammerMetrics class.
```
protocol7, line 33: Aside, we should really rename that class, Yammer hasn't been involved for some time now.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+### 1.1.0
+Add `folsom-semantic-metrics` module containing a [semantic-metrics][]-based implementation of folsom's `Metrics`.
```
protocol7: Nit: s/folsom's/Folsom's/

protocol7 (approved):
+1

mattnworb:
Looks like the build status is stuck in pending because the refactor to be a multimodule build broke coveralls somehow. I'll look into this.

folsom-semantic-metrics/src/main/java/com/spotify/folsom/client/SemanticFolsomMetrics.java
mattnworb: how about upgrading to the latest version of dropwizard-metrics and renaming the class to that? I can do that in another PR. 

Do you think its worth moving that class into its own module too?

CHANGELOG.md
mattnworb: fixed

mattnworb:
Closing as I think coveralls isn't updating the build status properly if the PR is from a fork.

PR #89, opened 2018-04-19
Specifiable Netty EventLoopGroup and executor

aquadrini:
I've added a few changes mainly regarding Netty behaviour:

- Added support to specify an existing EventLoopGroup to reuse.
- Added support for Epoll ELG and Channel.
- Allow to execute callbacks in same ELG threads, specifying a null executor in builder, for efficient execution of very short lived tasks.
- Dependencies upgrade.
- Reduce memory footprint via AtomicReferenceFieldUpdater usage

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
@@ -19,8 +19,6 @@ package com.spotify.folsom.client;
 import com.google.common.collect.Queues;
 import com.spotify.folsom.guava.HostAndPort;
 import com.spotify.futures.CompletableFutures;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 
 import com.spotify.folsom.AbstractRawMemcacheClient;
 import com.spotify.folsom.MemcacheClosedException;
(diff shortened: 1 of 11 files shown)


folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -69,27 +73,26 @@
...
   private final long timeoutMillis;
   private final int maxSetLength;
 
-  private final AtomicReference<String> disconnectReason = new AtomicReference<>(null);
+  private static final
+    AtomicReferenceFieldUpdater<DefaultRawMemcacheClient, String> DISCONNECT_REASON_UPDATER =
+          AtomicReferenceFieldUpdater.newUpdater(DefaultRawMemcacheClient.class,
+                  String.class, "disconnectReason");
+
+  private volatile String disconnectReason;
```
spkrka: Is this change really worth the added code complexity? The main purpose of `AtomicReferenceFieldUpdater` seems to be to reduce memory usage, but the number of active `DefaultRawMemcacheClient` is typically fairly small.

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -114,7 +119,7 @@
       decoder = new AsciiMemcacheDecoder(charset);
     }
 
-    final ChannelHandler initializer = new ChannelInitializer<Channel>() {
+    final ChannelHandler initializer = new ChannelInitializer() {
```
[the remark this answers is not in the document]
aquadrini: It is, sorry. It's just an IDE issue that slipped through. I've fixed it.

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
aquadrini: Hi! Thanks for your review! I've just added it to follow Netty style of using references, to reduce memory footprint as much as possible in every case. It's true that client instance count will be usually low, so it's ok for me if you prefer to rollback the change.

spkrka: Yes, I think we would prefer that. If you revert that, we will be happy to squash-merge and make a new release

aquadrini:
Great! I've just reverted the AtomicReferenceFieldUpdater changes. Thanks!

spkrka:
Thanks! I have released 1.1.0 now

Issue #91, opened 2018-09-04
Client failure after responding to too many requests where value is too large


morrifeldman:
I ran into an issue in production where we started getting `MemcacheOverloadedException`s and the client was never able to reconnect.  In our use case we are perhaps abusing the client with 1 to 2 requests per minute to cache values that are too large and after a few days the client crashes.  I think I've tracked the problem to the code [here](https://github.com/spotify/folsom/blob/master/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java#L216). When responding for a `VALUE_TO_LARGE` the pendingCounter is not decremented.  Looking at the outstandingRequests metric, I can see this counter get incremented for every request to cache a too large value.  Once the pending requests counter reaches the specified outstanding request limit (default 1000) the client fails and the application must be restarted.

The fix looks straight forward and I'll be happy to submit a pull request if you'd like.

spkrka:
Thanks for the report and PR! We'll merge either that one or #93 soon.

spkrka:
Should be fixed in release 1.1.1 now.

PR #92, opened 2018-09-04
Decrement pending request counter when handling VALUE_TOO_LARGE

morrifeldman:
Pending request counter is not decremented when handling VALUE_TOO_LARGE.  Fixing #91

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
@@ -214,6 +214,7 @@ public class DefaultRawMemcacheClient extends AbstractRawMemcacheClient {
       SetRequest setRequest = (SetRequest) request;
       byte[] value = setRequest.getValue();
       if (value.length > maxSetLength) {
+        pendingCounter.decrementAndGet();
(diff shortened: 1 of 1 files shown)


folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -214,6 +214,7 @@ private DefaultRawMemcacheClient(final HostAndPort address,
       SetRequest setRequest = (SetRequest) request;
       byte[] value = setRequest.getValue();
       if (value.length > maxSetLength) {
+        pendingCounter.decrementAndGet();
```
danielnorberg, line 217: Good catch. Seems to me like this length check should be done before incrementing the pending counter even, avoiding the need for this decrement.

spkrka:
There are two ways we can get VALUE_TOO_LARGE - either from within the client itself or if the server responds with that for a set operation. Is this bug only triggered by the former case?

I think I agree with @danielnorberg that it would be better to check before incrementing at all instead.

spkrka:
Here's an alternative PR, following @danielnorberg 's suggestion: https://github.com/spotify/folsom/pull/93

PR #104, opened 2018-11-09
Fix thread/FD leak

spkrka:


folsom/pom.xml
```diff
@@ -13,7 +13,7 @@
   <name>folsom</name>
 
   <properties>
-    <netty.version>4.1.23.Final</netty.version>
+    <netty.version>4.1.31.Final</netty.version>
```
spkrka, line 16: Probably not strictly needed

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -83,6 +85,14 @@
   private final int maxSetLength;
 
   private final AtomicReference<String> disconnectReason = new AtomicReference<>(null);
+  private static final ThreadFactory THREAD_FACTORY =
+      new DefaultThreadFactory(DefaultRawMemcacheClient.class, true);
+  private static final Supplier<EventLoopGroup> DEFAULT_EVENT_LOOP_GROUP =
```
spkrka, line 90: is it bad to reuse the same EventLoopGroup by default?

I tried shutting down the EventLoopGroup in the disconnect-method, but that stalled infinitely

spkrka:
@danielnorberg @protocol7

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
mattnworb: this might help: https://stackoverflow.com/questions/47134860/is-it-a-good-idea-to-use-the-same-netty-eventloopgroup-for-both-serverbootstrap

(I was curious about the same question)

spkrka:
Bug was introduced in https://github.com/spotify/folsom/commit/056ad841701c13ecaad22a3520d2bc3c1705395c

folsom/src/test/java/com/spotify/folsom/reconnect/ReconnectingClientTest.java
```diff
@@ -203,4 +211,44 @@ public void registerForConnectionChanges(ConnectionChangeListener listener) {
...
+    embeddedServer.start();
+    client.awaitFullyConnected(1, TimeUnit.MINUTES);
+  }
+
+  private long getFDCount() {
+    OperatingSystemMXBean os = ManagementFactory.getOperatingSystemMXBean();
+    if (os instanceof UnixOperatingSystemMXBean) {
+      return ((UnixOperatingSystemMXBean) os).getOpenFileDescriptorCount();
+    }
+    throw new UnsupportedOperationException("Unavailable");
```
mattnworb, line 252: you could also use one of the junit [Assume](https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html) methods to ignore the test if certain conditions are not met rather than throw an exception and fail the test

mattnworb (commented):
👍

folsom/src/test/java/com/spotify/folsom/reconnect/ReconnectingClientTest.java
spkrka: Good point! I should have done that - will tweak later.

PR #107, opened 2019-01-29
Support configurable batch size when creating clients

alexeyOnGitHub:
added `withRequestBatchSize(int)` method to MemcacheClientBuilder.

this allows fine-tuning performance for high-throughput services.
we have been using this fix in highly loaded services in prod env
for years (our local clone of this library).

extracted default batch size constant from BatchFlusher class
to Settings because it was not visible to code outside of that package
and it was required in MemcacheClientBuilder

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
+++ b/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
@@ -94,6 +94,7 @@ public class MemcacheClientBuilder<V> {
 
   private final List<HostAndPort> addresses = new ArrayList<>();
   private int maxOutstandingRequests = DEFAULT_MAX_OUTSTANDING;
+  private int batchSize = Settings.DEFAULT_BATCH_SIZE;
   private final Transcoder<V> valueTranscoder;
   private Metrics metrics = NoopMetrics.INSTANCE;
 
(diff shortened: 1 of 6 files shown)


folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -276,6 +277,28 @@ public MemcacheClientBuilder(final Transcoder<V> valueTranscoder) {
...
+   *
+   * <p>The optimal value should be matched to your workload and roughly the same value as your
+   * memcached server's `-R` argument, which defaults to 20.
+   *
+   * @param batchSize the maximum number of operations per batched client request. Default is
+   *     {@value Settings#DEFAULT_BATCH_SIZE}.
+   * @return itself
+   */
+  public MemcacheClientBuilder<V> withRequestBatchSize(final int batchSize) {
+    this.batchSize = batchSize;
```
spkrka, line 300: We should probably also sanity check the value here (must be >= 1 at least)

spkrka:
Thanks! This looks useful. I'd like @danielnorberg to take a look too though, since he wrote most of the current batching.

folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
alexeyOnGitHub: added param check

folsom/src/main/java/com/spotify/folsom/Settings.java
```diff
@@ -0,0 +1,9 @@
+package com.spotify.folsom;
+
+public class Settings {
+  /**
+   * Note that the default Memcached server batch size is 20. See Memcached server documentation on
+   * "-R" option.
+   */
+  public static final int DEFAULT_BATCH_SIZE = 64;
```
danielnorberg, line 8: Should this then default to `20` ?

folsom/src/main/java/com/spotify/folsom/client/BatchFlusher.java
```diff
@@ -64,8 +63,13 @@ public void run() {
         }
       };
 
+  /**
+   * Deprecated - use {@link BatchFlusher#BatchFlusher(Channel, int)} with explicit maxPending
+   * parameter instead.
+   */
+  @Deprecated
```
danielnorberg, line 70: No need to keep this constructor around, this class is not public anyway.

folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -276,6 +278,29 @@ public MemcacheClientBuilder(final Transcoder<V> valueTranscoder) {
     return this;
   }
 
+  /**
+   * Specify the maximum number of operations that will be batched together in one network request.
```
danielnorberg, line 282: This is slightly confusing. Each operation is still a separate memcache request. It's just that the requests are written to the socket in a fewer writes (sometimes a single). Would it make sense to clarify this?

danielnorberg:
Had some comments but looks mostly good to me.

spkrka (approved):

PR #108, opened 2019-01-29
more precise timeouts handling plus counting full request time

alexeyOnGitHub:
this change enables much more precise handling of request timeouts.

we have been using this in highly loaded services
with pretty low timeout setting (well under one second).

changed `TimeoutChecker` to use the time request was created.
changed `DefaultRawMemcacheClient` to poll for timeout every 10ms.
changed single timeout logging from error to info level. this helps with using the library in very high throughput services where you don't necessarily want to log every connection issue as an error.

added `createdNanos` to `Request` object

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
@@ -18,7 +18,6 @@ package com.spotify.folsom.client;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
-import static java.util.concurrent.TimeUnit.SECONDS;
 
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
@@ -72,9 +71,7 @@ public class DefaultRawMemcacheClient extends AbstractRawMemcacheClient {
 
   private static final AtomicInteger GLOBAL_CONNECTION_COUNT = new AtomicInteger();
   private static final double NANOS_IN_MILLIS = 1000000.0;
-  /**
-   * how often to check if the request has timed out
-   */
+  /** how often to check if the request has timed out */
   private static final int TIMEOUT_POLL_INTERVAL_MILLIS = 10;
 
   private final Logger log = LoggerFactory.getLogger(DefaultRawMemcacheClient.class);
--- a/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
@@ -71,6 +71,11 @@ import org.slf4j.LoggerFactory;
 public class DefaultRawMemcacheClient extends AbstractRawMemcacheClient {
 
   private static final AtomicInteger GLOBAL_CONNECTION_COUNT = new AtomicInteger();
+  private static final double NANOS_IN_MILLIS = 1000000.0;
+  /**
+   * how often to check if the request has timed out
+   */
+  private static final int TIMEOUT_POLL_INTERVAL_MILLIS = 10;
 
   private final Logger log = LoggerFactory.getLogger(DefaultRawMemcacheClient.class);
 
@@ -283,13 +288,12 @@ public class DefaultRawMemcacheClient extends AbstractRawMemcacheClient {
   private class ConnectionHandler extends ChannelDuplexHandler {
 
     private final Queue<Request<?>> outstanding = Queues.newArrayDeque();
-    private final TimeoutChecker<Request<?>> timeoutChecker =
+    private final TimeoutChecker timeoutChecker =
         TimeoutChecker.create(MILLISECONDS, timeoutMillis);
 
     private final Future<?> timeoutCheckTask;
 
     ConnectionHandler() {
-      final long pollIntervalMillis = Math.min(timeoutMillis, SECONDS.toMillis(1));
       timeoutCheckTask =
           channel
               .eventLoop()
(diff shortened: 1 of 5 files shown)


folsom/src/test/java/com/spotify/folsom/client/TimeoutCheckerTest.java
```diff
@@ -18,33 +18,72 @@
...
+    assertFalse(sut.check(request));
   }
 
   @Test
-  public void testNullNoTimeout() throws Exception {
-    assertFalse(sut.check("foo"));
-    assertFalse(sut.check(null));
+  public void testTimeout95ms() throws Exception {
+    request = new DummyRequest();
+    Thread.sleep(95);
```
mattnworb, line 43: a small suggestion: refactor `TimeoutChecker` to use a `java.time.Clock` instance for checking elapsed time instead, which you can then control with a fake/fixed Clock in the unit test, removing the need for `Thread.sleep` in tests.

(I'm also not sure if you can rely on a thread being slept for precisely 95 or 101 milliseconds)

spkrka: I was going to suggest something similar, but figured we could refactor that afterwards

alexeyOnGitHub: I like the idea of injecting a custom Clock into TimeoutChecker and using a FixedClock for testing. the good thing is that TimeoutChecker is an internal class and can be refactored later without breaking public API.

alas, I am not very familiar with Clock API so I cannot provide an immediate good-quality solution using it. my first glance suggests that there may be some issues with **nanoseconds precision** with Clock class.
e.g. this sample code shows very different values returned by `System.nanoTime()` and by `clock.instant.xxx`:
```    
    long createdNanos = System.nanoTime();
    Clock clock = Clock.systemUTC();
    long anotherNanos = TimeUnit.NANOSECONDS.convert(clock.instant().toEpochMilli(), TimeUnit.MILLISECONDS);
    System.out.println(createdNanos + " request nanos");
    System.out.println(anotherNanos + " request nanos - alternative");
```
output: 

> 1885962380543984 request nanos
> 1548885786068000000 request nanos - alternative

mattnworb: Good point - I forgot that Clock didn’t provide a similar method to nanoTime(). 👍

folsom/src/test/java/com/spotify/folsom/MemcacheClientBuilderTest.java
```diff
@@ -108,7 +108,7 @@ public void testOverloaded() throws Throwable {
     AsciiMemcacheClient<String> client =
         MemcacheClientBuilder.newStringClient()
             .withAddress(server.getHost(), server.getPort())
-            .withMaxOutstandingRequests(100)
+            .withMaxOutstandingRequests(10)
```
danielnorberg, line 111: spurious change?

folsom/src/main/java/com/spotify/folsom/client/Request.java
```diff
@@ -28,9 +28,11 @@
 
 public abstract class Request<V> extends CompletableFuture<V> {
   protected final byte[] key;
+  protected final long createdNanos;
```
danielnorberg, line 31: Why moving this timestamp to the `Request` instead of keeping it in `TimeoutChecker` ?

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -300,12 +301,13 @@ public int numActiveConnections() {
                       return;
                     }
                     if (timeoutChecker.check(head)) {
-                      log.error("Request timeout: {} {}", channel, head);
+                      double elapsedMs = timeoutChecker.elapsedNanos(head) / NANOS_IN_MILLIS;
```
danielnorberg, line 304: Could the `check()` method return the elapsed nanos directly instead of calculating elapses nanos twice?

Do we need this double division? Can we just log the nanos instead or go with a cheaper integer `TimeUnit.NANOSECONDS.toMillis()`?

folsom/src/main/java/com/spotify/folsom/client/Request.java
```diff
@@ -28,9 +28,11 @@
...
 
   protected Request(byte[] key) {
     this.key = key;
+    createdNanos = System.nanoTime();
```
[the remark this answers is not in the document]
spkrka: Thinking more about it - if someone creates a lot of requests in a batch, they will end up being queued, so the creation time of the request is not what we want.

Instead, I think the TimeoutChecker should keep state of when the head-request was first noticed. Which is what the code already does.

I would however be ok with checking for timeouts more frequently with the pollIntervalMillis change.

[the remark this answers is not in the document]
alexeyOnGitHub: I can see value in the current behavior (check when the request was first noticed).
at the same time -
from client point of view, if I define a timeout of 100ms and send a request which then spends 95 msec in a queue and then 10 msec in flight, I still want it to time out since it took longer than 100 msec overall.
we use tons of microservices and they all have specific SLAs. accumulating extra processing time during a request chain may lead to negative end-user experience. I would rather get a timeout after 100 msec from the moment when I first initialize the Memcached request. this timeout can serve as a signal that I may need to adjust my batch size, maybe prioritize local cache,  or find some other solution -  but it is a better outcome for me than requests taking longer than max configured time of 100 msec.

another option is to separate "max total time" and "max flight time" options, but that may be an overkill because many apps probably won't care about these two, and a few highly loaded apps will *probably* want the total time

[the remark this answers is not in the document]
spkrka: I think there are two separate use cases here:
1. detecting and recovering from broken connections (what's already in place, although parts of this PR could be used to improve it)
2. setting an upper limit of acceptable wait-time for requests.

Use case 2 could be somewhat implemented externally to folsom by adding timeouts (like this perhaps: http://iteratrlearning.com/java9/2016/09/13/java9-timeouts-completablefutures.html ) though that wouldn't have other nice features such as fast-failing the folsom requests if we're falling behind.

I would prefer to keep the existing functionality, but possibly add new functionality to timeout individual requests

alexeyOnGitHub:
alright, I submitted a separate PR to increase precision - https://github.com/spotify/folsom/pull/111
we can close this one.

PR #109, opened 2019-01-30
extracted yammer metrics to a separate module to limit dependencies

alexeyOnGitHub:
yammer metrics are now in a separate module because not everyone
would want to use them, and pulling in the tree of Yammer Metrics
dependencies may lead to unnecessary build complications
in consuming code.

I duplicated FakeRawMemcacheClient used in two sets of tests.
here duplication may be an acceptable solution, given the alternative
would be introducing "folsom-test" module with FakeRawMemcacheClient
class in it, which would be used from several places. but that module
would need a dependency on "folsom" module, while "folsom" module's
tests would need to access that dependency in a cycle.

a cleaner solution would be to still introduce "folsom-test" module
and move FakeRawMemcacheClient there, but then *also* move out
"ketama client", "round robin client", etc to separate modules - that
would allow avoiding dependency cycles. while that may be a good
long-term solution, I want to avoid the scope of changes in this branch,
where the goal is to remove "yammer core" dependencies tree
from the build, unless explicitly requested by the consumer.


spkrka:
Thanks! A somewhat simpler solution if you want to avoid the yammer dependencies is to mark it as `<scope>optional</scope>`. This would however be somewhat inconsistent with how the semantic-metrics module works.

We could also move `FakeRawMemcacheClient` into `src/main/java/` in a `.test` sub-package to avoid the duplication.

PR #114, opened 2019-02-06
Improve ttl logic

spkrka:


flozano:
This makes it consistent between ASCII and binary, but it won't fix the actual issue of timestamp not expiring when expected, right?

spkrka:
It also avoids converting from `ttl` to an expiration timestamp when `ttl < 30 days` (for both binary and ascii) so I think it should help your usecase too.
Perhaps the issue was misalignment in clocks between your client and server? If so, using ttls should be more preferable than timestamps.

flozano:
Actually the server and the client are in the same laptop in the cases it's failing, so I don't think it's due to misalignment at all...

protocol7:
+1, but maybe add a test for Utils

spkrka:
Added some tests!

spkrka:
@flozano yes, but perhaps your memcached server computes local time in seconds differently compared to how folsom does it in some way? One of them could be UTC, other one local timezone, or something similar.

Regardless, I think it's safer to send the expiration as TTL in seconds instead of as a timestamp.

flozano:
I agree that fixing the TTL inconsistency for the binary protocol is a good idea, thanks for the effort, the improvement in this PR is good overall.

I just wonder why it fails with timestamps. In my opinion, the original issue should be kept open - or a new, more specific one if you prefer - as I think the behaviour we're seeing is really unexpected.

With regard to timestamp offset, I think timestamps are number of seconds since Jan 1st 1970 00:00:00 UTC everywhere, timestamp should not influence...

spkrka:
Yes, once this is released, could you try to find some memcached docker image that can reproduce the problem? The existing new test case should hopefully be able to catch it.

If the problem only triggers when sending the expiration as a timestamp it would be harder to verify as you would need to have a test that waits for 30 days...

jorcasso:
Why is this conversion needed?
```
int expirationTime = (int) (System.currentTimeMillis() / 1000) + ttl;
```
If `ttl >= TTL_CUTOFF`, does not mean that `ttl` should be treated as Unix time?

spkrka:
According to the regular Memcached API, yes.
The Folsom API however, specifies it should be TTL in seconds.

jorcasso:
I see, that makes sense then. Thanks.

flozano:
Actually now this issue is almost un-testable for the original issue, as you said... It only happens with timestamp, so ... now I need to wait for 30 days for this to happen?

spkrka:
I'll think some more about this. Some ideas are:
* Add a separate API to trigger this type of expiration
* Make the cutoff configurable somehow
* Add tests that modify the system clock of the memcached container (not sure how easy this would be)

Though overall, do you need to test it? 30 days cache TTL is quite long and I can't think of any usecases with higher TTLs.
If you have such a usecase, it may be worth following up on, but if you don't, perhaps we can leave it for now.

PR #116, opened 2019-02-07
Fix problem with outstanding-requests metrics

spkrka:


folsom-semantic-metrics/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>com.spotify</groupId>
     <artifactId>folsom-parent</artifactId>
-    <version>1.3.1-SNAPSHOT</version>
+    <version>2.0.0-SNAPSHOT</version>
```
spkrka: Since I'm changing the interfaces perhaps I need to bump the major version? Not sure if this is overkill or not.

folsom-semantic-metrics/src/main/java/com/spotify/folsom/client/SemanticFolsomMetrics.java
```diff
@@ -70,6 +72,8 @@
   private final SemanticMetricRegistry registry;
   private final MetricId id;
 
+  private final Set<OutstandingRequestsGauge> gauges = Sets.newCopyOnWriteArraySet();
```
spkrka, line 75: We expect to read the gauge more frequently than we reconnect nodes

folsom/src/main/java/com/spotify/folsom/Metrics.java
```diff
@@ -38,6 +38,8 @@
    */
   void registerOutstandingRequestsGauge(OutstandingRequestsGauge gauge);
 
+  void unregisterOutstandingRequestsGauge(OutstandingRequestsGauge gauge);
```
spkrka, line 41: This technically breaks the API, but in practice only ones who implement the interface are shipped in packages that are versioned with the same number so probably ok anyway.

protocol7: Yeah, this seems okay without bumping major to me.

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -188,14 +190,16 @@ private DefaultRawMemcacheClient(
     this.address = address;
     this.executor = executor;
     this.timeoutMillis = timeoutMillis;
+    this.metrics = metrics;
     this.maxSetLength = maxSetLength;
     this.channel = checkNotNull(channel, "channel");
     this.flusher = new BatchFlusher(channel, batchSize);
     this.outstandingRequestLimit = outstandingRequestLimit;
+    this.pendingCounterLimit = Math.max(Integer.MAX_VALUE / 2, outstandingRequestLimit);
```
protocol7, line 204: See this code was moved from below, but still not sure why we use this max. If you know, maybe add a comment on it?

spkrka: I could explain here... but better to add comment - will do!

protocol7 (approved):

PR #117, opened 2019-02-07
Use lazy suppliers to ensure avoiding creating unnecessary executors

spkrka:


danielnorberg (commented):
These are already lazy though.

folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -103,7 +107,7 @@
 
   private int connections = 1;
   private boolean retry = true;
-  private Executor executor = DefaultExecutor.INSTANCE;
```
spkrka, line 106: @danielnorberg I think this line would make it non-lazy

danielnorberg: Indeed, so only change this line? The memoized supplier doesn’t seem like it would’ve avoided this issue.

spkrka: Yes, that would be an option too. I suppose this is more of a stylistic change. `Suppliers.memoize()` could be considered more obvious than the solution with a private static inner class.

danielnorberg: Yeah, I can buy that.

danielnorberg (approved):

PR #124, opened 2019-02-25
Implement deleteAll

spkrka:


folsom/src/main/java/com/spotify/folsom/MemcacheClient.java
```diff
@@ -49,6 +49,15 @@
    */
   CompletionStage<MemcacheStatus> delete(String key);
 
+  /**
+   * Delete the provided key on all memcached instances. This is typically only useful for a
+   * multi-instance setup (using Ketama).
+   *
+   * @param key Key, must not be null
+   * @return A future representing completion of the request
```
protocol7, line 57: Maybe clarify how this works with multiple servers, i.e. I assume that any failed requests will cause a failed future.

protocol7:
+1

PR #128, opened 2019-04-03
Implement getStats()

spkrka:


folsom/src/main/java/com/spotify/folsom/MemcachedStats.java
```diff
@@ -0,0 +1,15 @@
+package com.spotify.folsom;
```
protocol7, line 16: Missing copyright header

protocol7:
Needs a few more tests for the request classes and parsers.

folsom/src/main/java/com/spotify/folsom/MemcachedStats.java
spkrka: Will fix

spkrka:
Yes, some more unit tests might make sense, as opposed to only having integration tests

I will also try to prepare semantic metrics support (as a separate PR) before merging this, to be more certain this approach makes sense.

PR #135, opened 2019-06-14
Change OpCode to an enum

protocol7:
Move OpCode into an enum of byte values. This change is expected to be
backwards compatible, but would cause invalid responses to throw an
exception (for an unknown opcode).

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/client/OpCode.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/OpCode.java
@@ -16,75 +16,92 @@
 
 package com.spotify.folsom.client;
 
-public final class OpCode {
+import java.util.EnumSet;
 
-  private OpCode() {}
+public enum OpCode {
+  GET(0x00),
+  SET(0x01),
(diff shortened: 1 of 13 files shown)


folsom/src/main/java/com/spotify/folsom/client/OpCode.java
```diff
@@ -16,75 +16,93 @@
...
+
+  TAP_MUTATION(0x41),
+  TAP_DELETE(0x42),
+  TAP_FLUSH(0x43),
+  TAP_OPAQUE(0x44),
+  TAP_VBUCKET_SET(0x45),
+  TAP_CHECKPOINT_START(0x46),
+  TAP_CHECKPOINT_END(0x47);
+
+  private static final Map<Byte, OpCode> BY_VALUE;
```
spkrka: Since this is used in the critical path, can we instead use a primitive array?
`OpCode[] BY_VALUE = new OpCode[0x50]` or something like that

protocol7:
Offline +1:d by @spkrka

PR #136, opened 2019-06-15
Tracing

protocol7:
This PR is based on top of #135 so please review that first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,9 @@ API.
 ### Runtime dependencies
 
 * Netty 4
-* Yammer metrics
 * Google Guava
+* Yammer metrics (optional)
+* OpenCensus (optional)
 
 ### Usage
 
@@ -52,7 +53,15 @@ To import it with maven, use this:
       <version>1.6.0</version>
     </dependency>
 
-If you want to use one of the metric libraries, make sure you use the same version as the main artifact.
+    <!-- optional if you want to expose folsom tracing with OpenCensus -->
+    <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>folsom-opencensus</artifactId>
+      <version>1.6.0</version>
+    </dependency>
+
+If you want to use one of the metrics or tracing libraries, make sure you use the same version as
+the main artifact.
 
 We are using [semantic versioning](http://semver.org)
 
@@ -147,8 +156,24 @@ get a period of low cache hit ratio.
 
 #### Yammer metrics
 
-You can optionally choose to track performance using yammer metrics. You will need to include
-`folsom-yammer-metrics` dependency in this case.
+You can optionally choose to track performance using
+[Yammer metrics](https://metrics.dropwizard.io/4.0.0/).
+You will need to include the `folsom-yammer-metrics` dependency and initialize
+using MemcacheClientBuilder:
+
+```
+builder.withMetrics(new YammerMetrics(metricsRegistry));
+```
+
+#### OpenCensus tracing
+
+You can optionally use [OpenCensus](https://opencensus.io/) to trace Folsom operations.
+You will need to include the `folsom-opencensus` dependency and initialize tracing
+using MemcacheClientBuilder:
+
+```
+builder.withTracer(OpenCensus.tracer());
+```
 
 ### Building
 
--- a/folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensus.java
+++ b/folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensus.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2019 Spotify AB
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.spotify.folsom.opencensus;
+
+import com.spotify.folsom.Tracer;
+
+/** Starting point for the OpenCensus tracing implementation for Folsom */
+public class OpenCensus {
+
+  private static final boolean DEFAULT_INCLUDE_KEYS = true;
+  private static final boolean DEFAULT_INCLUDE_VALUES = false;
+
+  /**
+   * Get tracer, usually to be provided to {@link
+   * com.spotify.folsom.MemcacheClientBuilder#withTracer(Tracer)}.
+   */
+  public static Tracer tracer() {
+    return new OpenCensusTracer(DEFAULT_INCLUDE_KEYS, DEFAULT_INCLUDE_VALUES);
+  }
+
+  public static class Builder {
+    private boolean includeKeys = DEFAULT_INCLUDE_KEYS;
+    private boolean includeValues = DEFAULT_INCLUDE_VALUES;
+
+    /**
+     * If true, the key used for an operation will be included as a span attribute. Default to true.
+     */
+    public Builder withIncludeKeys(final boolean includeKeys) {
+      this.includeKeys = includeKeys;
+      return this;
+    }
+
+    /**
+     * If true, the value used for an operation will be included as a span attribute. Default to
+     * false.
+     *
+     * <p>Note that enabling tracing with large values may add excessive load on the tracing system.
+     */
+    public Builder withIncludeValues(final boolean includeValues) {
+      this.includeValues = includeValues;
+      return this;
+    }
+
+    public Tracer build() {
+      return new OpenCensusTracer(includeKeys, includeValues);
+    }
+  }
+}
--- a/folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusSpan.java
+++ b/folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusSpan.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2019 Spotify AB
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.spotify.folsom.opencensus;
+
+import static io.opencensus.trace.AttributeValue.longAttributeValue;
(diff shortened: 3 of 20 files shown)


protocol7:
@dmichel1 tracing PR as mentioned, would be awesome if you could take a look

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
```diff
@@ -0,0 +1,81 @@
...
+  private Span internalSpan(
+      final String name,
+      final CompletionStage<?> future,
+      final String operation,
+      final String key) {
+    requireNonNull(name);
+    requireNonNull(future);
+    requireNonNull(operation);
+
+    final io.opencensus.trace.Tracer tracer = Tracing.getTracer();
```
dmichel1: typically this is a static variable - https://github.com/census-instrumentation/opencensus-java/blob/master/examples/src/main/java/io/opencensus/examples/trace/MultiSpansTracing.java#L31

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
```diff
@@ -0,0 +1,81 @@
...
+
+    final io.opencensus.trace.Span ocSpan =
+        tracer.spanBuilder(name).setSpanKind(CLIENT).startSpan();
+
+    ocSpan.putAttribute("component", stringAttributeValue("folsom"));
+    ocSpan.putAttribute("peer.service", stringAttributeValue("memcache"));
+    ocSpan.putAttribute("operation", stringAttributeValue(operation));
+    if (key != null) {
```
dmichel1: should adding the key as an annotation be configurable? I'm thinking there might be scenarios where key is "sensitive."

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
```diff
@@ -0,0 +1,81 @@
...
+      final CompletionStage<?> future,
+      final String operation,
+      final String key) {
+    requireNonNull(name);
+    requireNonNull(future);
+    requireNonNull(operation);
+
+    final io.opencensus.trace.Tracer tracer = Tracing.getTracer();
+
+    final io.opencensus.trace.Span ocSpan =
```
dmichel1, line 79: this is the trickiest part of tracing especially with async operations happening...

with this implicit context propagation traces spans might not be linked when they go from a say an apollo thread pool -> memcached thread pool. 

if this is indeed the case here, the `span` can be added as parameter to these methods and passed in with the memcached operation. This would then become `tracer.spanBuilderWithExplicitParent()`



https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md#how-span-interacts-with-context

folsom/src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
```diff
@@ -90,6 +95,7 @@ public DefaultAsciiMemcacheClient(
     SetRequest request = SetRequest.casSet(byteKey, valueBytes, ttl, cas);
     CompletionStage<MemcacheStatus> future = rawMemcacheClient.send(request);
     metrics.measureSetFuture(future);
+    tracer.span("folsom.set", future, "set", key);
```
dmichel1: If we need to do explicit propagation this would have to look something like...
```java
public CompletionStage<MemcacheStatus> set(String key, V value, int ttl, long cas, Span span) {
...
tracer.span("folsom.set", future, "set", key, span);

```

dmichel1 (commented):
left a few comments.

can you add how to use this to the readme as well?

exciting!

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
protocol7: Yeah, this assumes that the context is propagated to the calling thread/gRPC Context. Once in Folsom-land, this should works as expected. But, this would be equally true for tracing in say Bigtable/Spanner, right (I've only looked at the Spanner client and it seems to use the same pattern as here)? If so, I believe we should make sure we propagate the tracing context correctly at least out of the box in Apollo, is this the case today?

folsom/src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
protocol7: I'd like to avoid polluting the Folsom API if at all possible :)

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
protocol7: I was thinking about this, which is why I opted not to include the value. But perhaps this is worth considering also with the key.

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
protocol7: Ah cool, will fix.

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
dmichel1: we should be propagating the context in apollo correctly as the traces to any gRPC based service (bigtable/pubsub/spanner) are linked already. have you tried testing this code with an apollo services yet?

folsom/src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
dmichel1: 👍 - agreed

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
dmichel1: knowing what key is potentially causing an issue could definitely be valuable during troubleshooting so I can see the benefit of leaving it here/enabled by default...

what about adding another attribute for the size of the payload being sent to memcached? that could also be valuable during troubleshooting

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
protocol7: I haven't yet, but got a service hopefully lined up

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
protocol7: Great suggestions! Pushed an update for addressing both of these. There's now a builder to configure if you want keys/values included in traces. Also added value size.

protocol7:
Fix pushed which addresses comments so far, including an updated README

README.md
```diff
@@ -52,7 +53,15 @@ To import it with maven, use this:
       <version>1.6.0</version>
     </dependency>
 
-If you want to use one of the metric libraries, make sure you use the same version as the main artifact.
+    <!-- optional if you want to expose folsom metrics with yammer -->
```
dmichel1: `s/folsom metrics with yammer/traces with OpenCensus/`

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusSpan.java
```diff
@@ -0,0 +1,63 @@
...
+    this.span = span;
+    this.includeValues = includeValues;
+  }
+
+  @Override
+  public Span value(final byte[] value) {
+    if (includeValues) {
+      span.putAttribute("value_hex", stringAttributeValue(HEX.encode(value)));
+    }
+    span.putAttribute("value_size", longAttributeValue(value.length));
```
dmichel1: include the unit here. e.g. `value_size_bytes`

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensus.java
```diff
@@ -0,0 +1,60 @@
...
+
+package com.spotify.folsom.opencensus;
+
+import com.spotify.folsom.Tracer;
+
+/** Starting point for the OpenCensus tracing implementation for Folsom */
+public class OpenCensus {
+
+  private static final boolean DEFAULT_INCLUDE_KEYS = true;
+  private static final boolean DEFAULT_INCLUDE_VALUES = false;
```
dmichel1, line 25: I'm a bit concerned about someone enabling the value attribute. It could be a really big payload. I think since folks have the ability to see the key if they really care about the value they can fetch it themselves.

I would be in favor of removing the value attribute entirely.

dmichel1 (approved):
Couple more comments and then let's try this out.

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
hexedpackets: I think this pattern is generally used for many libraries, Bigtable does it similarly. It may be useful to allow passing in the parent span instead of solely relying on `getCurrentSpan` - that way if the calling code is doing something unusual there's a way to keep the context correctly linked.

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusSpan.java
protocol7: Fixed

README.md
protocol7: Fixed

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensus.java
protocol7: I think it's fine as long as this is default to off, but I'll add a note about this in the documentation.

folsom-opencensus/src/main/java/com/spotify/folsom/opencensus/OpenCensusTracer.java
protocol7: @hexedpackets do you have a reference handy to how this is done in Bigtable, would be curious as to how they opted to design it? I tried to find, but not sure I'm looking in the right place.

For now, I would not want to change the Folsom API in order to support tracing, but if we see more requests from users on setting explicit parent spans I think we should look closer at how to do this in the future.

hexedpackets: That makes sense, I think the use case of needing to pass in the parent span should be pretty rare.

Bigtable [creates a span when the operation is instantiated](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java#L149) and then closes it in a callback [after the operation ends](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java#L161). I don't think this is passed around, but various functions in the class add tags and annotations to the span when they're run.

PR #141, opened 2019-07-15
Bump semantic-metrics version

mziccard:


spkrka:
Dependency diff for the folsom-semantic-metrics module:

Before:
```
 +- com.spotify.metrics:semantic-metrics-core:jar:0.10.0:compile
 |  +- com.spotify.metrics:semantic-metrics-api:jar:0.10.0:compile
 |  +- com.codahale.metrics:metrics-core:jar:3.0.2:compile
 |  \- com.codahale.metrics:metrics-jvm:jar:3.0.2:compile
```

After:
```
 +- com.spotify.metrics:semantic-metrics-core:jar:1.0.2:compile
 |  +- com.spotify.metrics:semantic-metrics-api:jar:1.0.2:compile
 |  +- io.dropwizard.metrics:metrics-core:jar:4.0.2:compile
 |  \- io.dropwizard.metrics:metrics-jvm:jar:4.0.2:compile
```

Does this warrant a major version bump for folsom itself? What do you think @protocol7 ?

spkrka:
Compatibility report: http://htmlpreview.github.io/?https://github.com/spotify/folsom/blob/krka/compat_report/compat_report.html

spkrka:
Metrics-core changes look pretty benign to me - probably ok to merge without bumping major

protocol7:
We've been bitten by this in Apollo as Metrics moves things around in 4.x. So I'd say this would require a major version bump in Folsom. But, maybe we should rather move out the semantic-metrics module to it's own repo as I suspect it's rather rarely used outside of Spotify? Would have the downside with no longer matching versions.

protocol7:
Discussed this offline, will merge this with #142 (which bumps guava) and do a major version bump after that.

protocol7:
+1

davidpoblador:
LGTM

protocol7:
I'll retract the above after having read up on semver which clearly says that cases like this should not cause a major version bump (https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api, https://github.com/semver/semver/issues/148).

PR #142, opened 2019-08-05
Bump dependencies

protocol7:


davidpoblador:
LGTM

folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -71,31 +71,34 @@
...
-                  ForkJoinPool.defaultForkJoinWorkerThreadFactory,
-                  new UncaughtExceptionHandler(),
-                  true));
+              () ->
+                  new ForkJoinPool(
+                      Runtime.getRuntime().availableProcessors(),
+                      ForkJoinPool.defaultForkJoinWorkerThreadFactory,
+                      new UncaughtExceptionHandler(),
+                      true))
+          ::get;
```
mbruggmann, line 80: Guess we could type the `Supplier` field and constants as the Guava one to avoid the `::get` calls here but I'm not sure if that's better.

mbruggmann (approved):

folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
protocol7: Yeah, I was considering the same. But in general I'd like to lessen our use of Guava in Folsom, and that would be a step in the wrong direction. I was also considering adding a local memoize function instead, using only java.util.function.Supplier, but decided to rather do so in a future PR.

mbruggmann: Makes sense. 👍

PR #143, opened 2019-09-26
Add utility function to check if an executor is direct or not

krka:
Don't move futures to different executor if it's a direct executor

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
@@ -83,6 +83,7 @@ public class DefaultRawMemcacheClient extends AbstractRawMemcacheClient {
   private final BatchFlusher flusher;
   private final HostAndPort address;
   private final Executor executor;
+  private final boolean isDirectExecutor;
   private final long timeoutMillis;
   private final Metrics metrics;
   private final int maxSetLength;
@@ -194,6 +195,7 @@ public class DefaultRawMemcacheClient extends AbstractRawMemcacheClient {
       int maxSetLength) {
     this.address = address;
     this.executor = executor;
+    this.isDirectExecutor = executor == null || Utils.isDirectExecutor(executor);
     this.timeoutMillis = timeoutMillis;
     this.metrics = metrics;
     this.maxSetLength = maxSetLength;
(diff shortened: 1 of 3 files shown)


folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -194,6 +195,7 @@ private DefaultRawMemcacheClient(
       int maxSetLength) {
     this.address = address;
     this.executor = executor;
+    this.isDirectExecutor = executor == null || Utils.isDirectExecutor(executor);
```
protocol7, line 198: Nitpicking here, but given isDirectExecutor really meaning "is a direct executor or null", how about rather naming it something like "wrapWithExecutor"? (I'm terrible at naming so I'm sure an even better name would be available)

folsom/src/main/java/com/spotify/folsom/client/Utils.java
```diff
@@ -72,4 +74,43 @@ public static Throwable unwrap(final Throwable e) {
...
+    requireNonNull(executor);
+    RunnableWithThread runnableWithThread = new RunnableWithThread();
+    try {
+      executor.execute(runnableWithThread);
+
+      // We have the following cases
+      // 1) It is a direct executor, so runnableWithThread.thread will be set to the current thread.
+      //    We will correctly return true
+      // 2) It is not a direct executor and runnableWithThread.thread is still null when we check
+      // it.
```
protocol7, line 95: Nitpicking: wrong indentation

folsom/src/main/java/com/spotify/folsom/client/Utils.java
```diff
@@ -72,4 +74,43 @@ public static Throwable unwrap(final Throwable e) {
...
+      //    We correctly return false
+      // 3) It is not a direct executor but the runnableWithThread.thread somehow managed to get set
+      //    before we check it. In any case, it can't be referencing the same thread we are in
+      //    We correctly return false
+
+      return runnableWithThread.thread == Thread.currentThread();
+    } catch (Exception e) {
+      // If the executor throws any exception, it can not be a direct executor
```
protocol7, line 103: Is this as safe assumption? Looking at the Guava DirectExecutorService, it can throw exceptions here, but only if the executor is shutting down, which is probably fine either way. But, would we want to assume other direct executors would behave the same way?

spkrka: No, probably not a safe assumption. It depends a bit on what the function is used for. If it's only for optimization purposes and sanity checks it may still be good enough

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
spkrka: Sure, that's a good idea

folsom/src/main/java/com/spotify/folsom/client/Utils.java
spkrka: Oops

PR #152, opened 2020-03-26
Flags support

paralainer:
Added support of memcached flags

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/MemcacheClient.java
+++ b/folsom/src/main/java/com/spotify/folsom/MemcacheClient.java
@@ -16,8 +16,6 @@
 package com.spotify.folsom;
 
 import com.spotify.folsom.client.Flags;
-
-import java.util.BitSet;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletionStage;
@@ -186,7 +184,8 @@ public interface MemcacheClient<V> extends ObservableClient {
    *
    * @param keys Keys, must not be null, nor must any key in the list
    * @return A future representing completion of the request, with the values, including the CAS
-   *     value and flags. Any non existing values will be null. Order will be maintained from the input keys
+   *     value and flags. Any non existing values will be null. Order will be maintained from the
+   *     input keys
    */
   CompletionStage<List<GetResult<V>>> casGet(List<String> keys);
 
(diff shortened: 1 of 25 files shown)


folsom/src/main/java/com/spotify/folsom/MemcacheClient.java
```diff
@@ -104,14 +153,25 @@
...
   CompletionStage<GetResult<V>> casGet(String key);
 
+  /**
+   * Get the value for the provided key, including the CAS value and flags
+   *
+   * @param key First key, must not be null
+   * @return A future representing completion of the request, with the value, including the CAS
+   *     value and flags, or null if the value does not exists.
+   */
+  default CompletionStage<GetResult<V>> getWithFlags(String key) {
```
paralainer, line 169: This is basically a copy of `casGet`, it looks better when we have a method that explicitly says that it returns flags as well.

folsom/src/main/java/com/spotify/folsom/client/ascii/AsciiMemcacheDecoder.java
```diff
@@ -179,6 +183,7 @@ protected void decode(final ChannelHandlerContext ctx, final ByteBuf buf, final
             this.value = new byte[size];
             this.valueOffset = 0;
             this.cas = cas;
+            this.flags = new Flags(flags);
```
spkrka, line 186: If we store this as a primitive int, we can avoid some unnecessary garbage creation

folsom/src/main/java/com/spotify/folsom/client/binary/BinaryMemcacheDecoder.java
```diff
@@ -66,7 +67,12 @@ protected void decode(final ChannelHandlerContext ctx, final ByteBuf buf, final
         return;
       }
 
-      buf.skipBytes(extrasLength);
+      Flags flags = Flags.DEFAULT;
+      if (extrasLength > 0) {
+        flags = new Flags(buf.readInt());
```
spkrka, line 72: use primitive instead of allocating?

spkrka:
I think if we want to introduce flags, we should either do it high level or low level:

High-level: Introduce a proper enumset of valid flags that are supported
Low-level: use a primitive int that works with the API

Right now this PR is mostly low-level, but with an int-wrapper class. I think we can remove that wrapper class.

paralainer:
> High-level

I don't think it makes sense as it is pretty much an arbitrary value and it's up to consumer to decide which flags they need.

> Right now this PR is mostly low-level, but with an int-wrapper class

To be honest the only reason I introduced Flags class is that `int` interfere with other parameters which are `long`, it broke some tests and will most likely break some consumers code if they rely on automatic `int -> long` conversion.
Other approach could be instead of reusing the same method names create new methods with `withFlags` suffix, like `setWithFlags`. 

On the point of allocations, if people won't use flags they won't have allocations as it will use the `Flags.DEFAULT` which is static (I can make an optimization on response side as well, so if flags are 0 we use `Flags.DEFAULT`).

Having `Flags` class we can implement some convenience methods in it later on, which can help with bitmasks for instance. Also if you use this library from Kotlin it make it easier to build extension methods.

PR #168, opened 2021-06-02
deprecate withRequestBatchSize

danielnorberg:
The `withRequestBatchSize` configuration property documentation is currently incorrect and misleading.

* Deprecate `withRequestBatchSize`.
* Introduce `withEventLoopThreadFlushMaxBatchSize` to cover the special
  use case served by `withRequestBatchSize`.

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
+++ b/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
@@ -21,7 +21,6 @@ import static com.google.common.base.Preconditions.checkState;
 import static com.spotify.folsom.client.MemcacheEncoder.MAX_KEY_LEN;
 import static java.util.Objects.requireNonNull;
 
-import com.google.common.annotations.Beta;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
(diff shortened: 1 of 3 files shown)


danielnorberg:
@spkrka ptal

folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -332,13 +365,15 @@ private void updateResolver() {
...
-   *     {@value Settings#DEFAULT_BATCH_SIZE}.
+   * @param eventLoopThreadFlushMaxBatchSize the maximum number of requests that will be written to
+   *     a connection in a single operation when sending requests on the {@link EventLoopGroup}
+   *     thread that is handling the connection. Default is {@value Settings#DEFAULT_BATCH_SIZE}.
    * @return itself
    */
-  public MemcacheClientBuilder<V> withRequestBatchSize(final int batchSize) {
-    checkArgument(batchSize > 0, "batch size must be > 0");
-    this.batchSize = batchSize;
+  public MemcacheClientBuilder<V> withEventLoopThreadFlushMaxBatchSize(
```
spkrka, line 373: I don't think we can rename the method without breaking the API

danielnorberg: `withRequestBatchSize` is still there. It just looks like it was renamed in the diff.

spkrka (approved):

danielnorberg:
🙏

PR #171, opened 2021-07-02
com.spotify:dns 3.2.2 which uses dnsjava 3.0.2

klaraward:
See internal PR for some context - https://ghe.spotify.net/java/bom/pull/538

### DIFF
--- a/folsom/pom.xml
+++ b/folsom/pom.xml
@@ -73,7 +73,7 @@
     </dependency>
     <dependency>
       <groupId>junit</groupId>
-      <artifacftId>junit</artifactId>
+      <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/folsom/pom.xml
+++ b/folsom/pom.xml
@@ -50,7 +50,7 @@
     <dependency>
       <groupId>com.spotify</groupId>
       <artifactId>dns</artifactId>
-      <version>3.3.1</version>
+      <version>3.3.2</version>
     </dependency>
     <dependency>
       <groupId>commons-lang</groupId>
--- a/folsom/pom.xml
+++ b/folsom/pom.xml
@@ -50,7 +50,7 @@
     <dependency>
       <groupId>com.spotify</groupId>
       <artifactId>dns</artifactId>
-      <version>3.2.2</version>
+      <version>3.3.1</version>
     </dependency>
     <dependency>
       <groupId>commons-lang</groupId>
--- a/folsom/pom.xml
+++ b/folsom/pom.xml
@@ -50,7 +50,7 @@
     <dependency>
       <groupId>com.spotify</groupId>
       <artifactId>dns</artifactId>
-      <version>3.1.4</version>
+      <version>3.2.2</version>
     </dependency>
     <dependency>
       <groupId>commons-lang</groupId>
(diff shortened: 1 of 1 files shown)


folsom/pom.xml
```diff
@@ -73,7 +73,7 @@
     </dependency>
     <dependency>
       <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
+      <artifacftId>junit</artifactId>
```
spkrka: This doesn't look correct

protocol7:
```suggestion
      <artifactId>junit</artifactId>
```

protocol7 (approved):

protocol7:
@klaraward seems like com.spotify:dns:3.2.2 was built with Java 11, while Folsom requires 8. Could we build com.spotify:dns with 8 as well?

klaraward:
@fdfzcq ?

fdfzcq:
@klaraward @protocol7 I've changed maven plugin version from 9 to 8, will it do? https://github.com/spotify/dns-java/pull/47

protocol7:
@fdfzcq yes, I think that would address this issue. Would you plan to release an update for com.spotify:dns with this fix?

fdfzcq:
@protocol7 yes we plan to merge and release that PR today

fdfzcq:
@protocol7 we merged the PR just to realize jdk 8 is not compatible with the latest `spotify/dns`. We will switch it back to jdk 9 for now and can look into moving it to jdk 8 (or multiple versions) later.

protocol7:
@fdfzcq Thanks for the update.

fdfzcq:
@klaraward @protocol7 we have now released newest version compiled with jdk 8: https://github.com/spotify/dns-java/releases/tag/v3.3.1

folsom/pom.xml
```diff
@@ -50,7 +50,7 @@
     <dependency>
       <groupId>com.spotify</groupId>
       <artifactId>dns</artifactId>
-      <version>3.1.4</version>
+      <version>3.2.2</version>
```
protocol7:
```suggestion
      <version>3.3.1</version>
```

protocol7:
@fdfzcq https://github.com/spotify/dns-java/pull/47 is a non-backwards compatible change in that it adds a new method to DnsSrvResolver. I don't know how common it would be to implement this among users of dns-java, but I could imagine it would be somewhat frequently implemented in tests. We can easily fix this in Folsom, but I would worry about rolling it, e.g. across the Spotify fleet.

fdfzcq:
@protocol7 right, erik actually pointed out it which I thought I fixed but didn't https://github.com/spotify/dns-java/pull/47/files#r711963085 We would like to eventually move all users to use `resolveAsync` but for now this shouldn't break them.

fdfzcq:
@protocol7 now third time charm :sweat_smile: https://github.com/spotify/dns-java/releases/tag/v3.3.2

folsom/pom.xml
```diff
@@ -50,7 +50,7 @@
     <dependency>
       <groupId>com.spotify</groupId>
       <artifactId>dns</artifactId>
-      <version>3.1.4</version>
+      <version>3.3.1</version>
```
protocol7:
```suggestion
      <version>3.3.2</version>
```

protocol7:
@fdfzcq that did indeed do the trick, thanks!

PR #182, opened 2021-09-28
Fix bug where unregister listener delegates to register listener

spkrka:


protocol7:
Could we cover this by tests?

protocol7 (approved):

PR #187, opened 2021-11-08
Implement flags support

mszabo-wikia:
Add flags support based on the existing work in https://github.com/spotify/folsom/pull/152, but using a primitive int rather than a wrapper type as suggested in the previous discussion.


folsom/src/main/java/com/spotify/folsom/MemcacheClient.java
```diff
@@ -106,14 +152,25 @@
...
   CompletionStage<GetResult<V>> casGet(String key);
 
+  /**
+   * Get the value for the provided key, including the CAS value and flags
+   *
+   * @param key First key, must not be null
+   * @return A future representing completion of the request, with the value, including the CAS
+   *     value and flags, or null if the value does not exists.
+   */
+  default CompletionStage<GetResult<V>> getWithFlags(String key) {
```
spkrka: Do we need to add this new method?

folsom/src/main/java/com/spotify/folsom/GetResult.java
```diff
@@ -18,19 +18,23 @@
...
 
-  private GetResult(final V value, final long cas) {
+  private GetResult(final V value, final long cas, final int flags) {
     this.value = requireNonNull(value);
     this.cas = cas;
+    this.flags = flags;
   }
 
-  public static <V> GetResult<V> success(final V value, final long cas) {
+  public static <V> GetResult<V> success(final V value, final long cas, final int flags) {
```
spkrka, line 35: Keep old method for compatibility

folsom-semantic-metrics/src/test/java/com/spotify/folsom/client/SemanticFolsomMetricsTest.java
```diff
@@ -93,7 +93,8 @@ public void testMultigetSuccess() {
     metrics.measureMultigetFuture(
         completedFuture(
             ImmutableList.of(
-                GetResult.success(new byte[] {1}, 0), GetResult.success(new byte[] {1}, 0))));
+                GetResult.success(new byte[] {1}, 0, 0),
```
spkrka: If the old success method is re-added, we can keep these tests unchanged I think

folsom/src/main/java/com/spotify/folsom/MemcacheClient.java
```diff
@@ -44,6 +54,18 @@
...
+   * Compare and set a key in memcache to the provided value, with the specified TTL
+   *
+   * @param key The key, must not be null
+   * @param value The value, must not be null
+   * @param ttl The TTL in seconds
+   * @param cas The CAS value, must match the value on the server for the set to go through
+   * @param flags Memcached flags
+   * @return A future representing completion of the request.
+   */
+  CompletionStage<MemcacheStatus> set(String key, V value, int ttl, long cas, int flags);
```
spkrka: Maybe we need to make all new interface methods `default` to avoid breaking compatibility if external code is implementing these interfaces.

In this case, the default method could just call the other set method

folsom/src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
```diff
@@ -161,11 +185,20 @@ public DefaultAsciiMemcacheClient(
 
   @Override
   public CompletionStage<MemcacheStatus> replace(String key, V value, int ttl) {
+    return replace(key, value, ttl, 0);
```
spkrka, line 188: If we do my other suggestion about making these interface methods `default`, we should instead delegate to an internal private method to avoid infinite recursion

folsom/src/main/java/com/spotify/folsom/client/binary/BinaryMemcacheDecoder.java
```diff
@@ -66,7 +66,12 @@ protected void decode(final ChannelHandlerContext ctx, final ByteBuf buf, final
         return;
       }
 
-      buf.skipBytes(extrasLength);
+      int flags = 0;
+      if (extrasLength > 0) {
```
spkrka: Do we need to check that extrasLength >= 4 here?

folsom/src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
mszabo-wikia: Sorry, I'm a bit confused about this one—wouldn't it be safe to make this call here since `DefaultAsciiMemcacheClient` overloads the default implementation?

mszabo-wikia:
@spkrka Thanks for the quick response! I've made some adjustments and also resolved formatting errors.

folsom-semantic-metrics/src/test/java/com/spotify/folsom/client/SemanticFolsomMetricsTest.java
```diff
@@ -52,7 +52,7 @@ public void testGetMiss() {
 
   @Test
   public void testGetHit() {
-    metrics.measureGetFuture(completedFuture(GetResult.success(new byte[] {}, 0)));
+    metrics.measureGetFuture(completedFuture(GetResult.success(new byte[] {}, 0, 0)));
```
spkrka:
```suggestion
    metrics.measureGetFuture(completedFuture(GetResult.success(new byte[] {}, 0)));
```

folsom/src/main/java/com/spotify/folsom/client/ascii/DefaultAsciiMemcacheClient.java
spkrka: Hm, yes I think you are right.

folsom/src/main/java/com/spotify/folsom/BinaryMemcacheClient.java
```diff
@@ -51,6 +64,19 @@
...
+   * memcache
+   *
+   * @param key The key, must not be null
+   * @param value The value, must not be null
+   * @param ttl The TTL in seconds
+   * @param cas The CAS value, must match the value on the server for the set to go through
+   * @param flags Memcached flags
+   * @return A future representing completion of the request, containing the new CAS value
+   */
+  CompletionStage<MemcacheStatus> replace(String key, V value, int ttl, long cas, int flags);
```
spkrka: add default implementation

folsom/src/main/java/com/spotify/folsom/BinaryMemcacheClient.java
```diff
@@ -39,6 +39,19 @@
...
+   * memcache
+   *
+   * @param key The key, must not be null
+   * @param value The value, must not be null
+   * @param ttl The TTL in seconds
+   * @param cas The CAS value, must match the value on the server for the set to go through
+   * @param flags Memcached flags
+   * @return A future representing completion of the request, containing the new CAS value
+   */
+  CompletionStage<MemcacheStatus> add(String key, V value, int ttl, long cas, int flags);
```
spkrka: add default implementation

folsom/src/main/java/com/spotify/folsom/client/ascii/SetRequest.java
```diff
@@ -108,6 +113,10 @@ private static ByteBuf toBufferWithValueAndNewLine(
     return buffer;
   }
 
+  private static byte[] toFlags(int flags) {
+    return (" " + flags + " ").getBytes(US_ASCII);
```
spkrka, line 123: Nit: If the set of valid flag combinations is small enough, we could make a lookup table of flags -> byte[]

But at the very least we should optimize the case where flags == 0 to use a constant

folsom/src/test/java/com/spotify/folsom/GetResultTest.java
```diff
@@ -24,16 +24,16 @@
 
   @Test(expected = NullPointerException.class)
   public void testCstrNullValue() {
-    GetResult.success(null, 123);
+    GetResult.success(null, 123, 0);
```
spkrka: revert this line

folsom/src/test/java/com/spotify/folsom/GetResultTest.java
```diff
@@ -24,16 +24,16 @@
...
   }
 
   @Test
   public void testEqualsHashCode() {
-    GetResult<Long> result1 = GetResult.success(123L, 456);
-    GetResult<Long> result2 = GetResult.success(123L, 456);
-    GetResult<Long> result3 = GetResult.success(999L, 999);
-    GetResult<Long> result4 = GetResult.success(123L, 999);
-    GetResult<Long> result5 = GetResult.success(999L, 456);
+    GetResult<Long> result1 = GetResult.success(123L, 456, 0);
```
spkrka: revert this line

folsom/src/test/java/com/spotify/folsom/retry/RetryingClientTest.java
```diff
@@ -44,7 +44,8 @@ public void testSimple() throws Exception {
     RawMemcacheClient delegate = mock(RawMemcacheClient.class);
 
     MemcacheClosedException ex = new MemcacheClosedException("reason");
-    GetResult<byte[]> result = GetResult.success(StringTranscoder.UTF8_INSTANCE.encode("bar"), 123);
+    GetResult<byte[]> result =
```
spkrka: revert this line

folsom/src/main/java/com/spotify/folsom/client/ascii/SetRequest.java
mszabo-wikia: Good point! I'm not sure if we should limit the available flags on the folsom side - wrappers may choose to utilize flags to denote serialization format or compression type used for a given value, so standardizing them on the client level might not end up being flexible enough. What do you think?

folsom/src/main/java/com/spotify/folsom/client/ascii/SetRequest.java
```diff
@@ -40,39 +41,46 @@
     CMD.put(Operation.APPEND, "append ".getBytes(US_ASCII));
     CMD.put(Operation.PREPEND, "prepend ".getBytes(US_ASCII));
     CMD.put(Operation.CAS, "cas ".getBytes(US_ASCII));
-  }
 
-  private static final byte[] FLAGS = " 0 ".getBytes(US_ASCII);
+    NO_FLAGS = "0".getBytes(US_ASCII);
```
protocol7: Does this need to go in the `static {}` block.

mszabo-wikia: Not really - I will move it to the same line as the declaration.

folsom/src/main/java/com/spotify/folsom/client/ascii/SetRequest.java
```diff
@@ -31,6 +31,7 @@
     implements com.spotify.folsom.client.SetRequest {
 
   private static final EnumMap<Operation, byte[]> CMD;
+  private static final byte[] NO_FLAGS = "0".getBytes(US_ASCII);
```
spkrka: This should be `" 0 "` I think

spkrka:
```suggestion
  private static final byte[] NO_FLAGS = " 0 ".getBytes(US_ASCII);
```

mszabo-wikia: Good catch, thanks!

spkrka:
Thanks for the PR! Merged and released as 1.10.0

mszabo-wikia:
Thanks!

PR #189, opened 2022-01-10
Move duplicate method to the Request interface

borislopezaraoz:
This change allows to easily duplicate any request.

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/client/Request.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/Request.java
@@ -18,5 +18,7 @@ public interface Request<V> {
 
   void handle(Object msg, HostAndPort address) throws IOException;
 
-  Request<V> duplicate();
+  default Request<V> duplicate() {
+    throw new RuntimeException("duplicate not implemented");
+  }
 }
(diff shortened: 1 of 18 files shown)


folsom/src/main/java/com/spotify/folsom/client/Request.java
```diff
@@ -17,4 +17,6 @@
   ByteBuf writeRequest(ByteBufAllocator alloc, ByteBuffer workingBuffer);
 
   void handle(Object msg, HostAndPort address) throws IOException;
+
+  Request<V> duplicate();
```
spkrka: I guess this needs to have a default implementation (that throws an exception) to avoid breaking binary compatibility

folsom/src/test/java/com/spotify/folsom/client/DefaultRawMemcacheClientTest.java
```diff
@@ -155,6 +155,11 @@ public ByteBuf writeRequest(ByteBufAllocator alloc, ByteBuffer dst) {
                   dst.put(NEWLINE_BYTES);
                   return toBuffer(alloc, dst);
                 }
+
+                @Override
+                public Request<String> duplicate() {
+                  return this;
```
spkrka: should instead throw if you don't expect this to be called in the test

javierarilos (approved):
LGTM

spkrka (approved):

PR #194, opened 2022-03-02
feat: support generic Serializable type on client

liufuyang:
# Support generic Serializable type on client

Current implementation is not that convenient when creating a client with `newSerializableObjectClient()` as it returns a `MemcacheClient<Serializable>`, which gives the following error when a user trying to use it with a generic type
image

So one has to write it as this to make it compile:
```java
  public record Student(String name, int age) implements Serializable { }

  public static void main(String[] args) throws Exception {
    MemcacheClient<Serializable> client =
        MemcacheClientBuilder.<Student>newSerializableObjectClient()
        .withAddress("localhost")
        .connectAscii();
    // make it wait until the client has connected to the server
    ConnectFuture.connectFuture(client).toCompletableFuture().get();

    client.set("key", new Student("fuyang", 28), 10000).toCompletableFuture().get();
    Student value = (Student) client.get("key").toCompletableFuture().get();
  }
```

This proposed change let `newSerializableObjectClient()` returns a `MemcacheClient<T extends Serializable>` instead, making the user much easier to use a client with a type that is implemented Serializable.

So after this change, a user can do:
```java
  public record Student(String name, int age) implements Serializable { }

  public static void main(String[] args) throws Exception {
    MemcacheClient<Student> client =
        MemcacheClientBuilder.<Student>newSerializableObjectClient()
        .withAddress("localhost")
        .connectAscii();
    // make it wait until the client has connected to the server
    ConnectFuture.connectFuture(client).toCompletableFuture().get();

    client.set("key", new Student("fuyang", 28), 10000).toCompletableFuture().get();
    Student value = client.get("key").toCompletableFuture().get();
  }
```

Plus testcontainers is updated to allow tests to run nicely on M1 machines.

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
+++ b/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
@@ -59,7 +59,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
-public class MemcacheClientBuilder<V extends Serializable> {
+public class MemcacheClientBuilder<V> {
 
   private static final int DEFAULT_MAX_SET_LENGTH = 1024 * 1024;
 
--- a/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
+++ b/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
@@ -152,7 +152,7 @@ public class MemcacheClientBuilder<V> {
    * @return The builder
    */
   public static <T extends Serializable> MemcacheClientBuilder<T> newSerializableObjectClient() {
-    return new MemcacheClientBuilder<>(SerializableObjectTranscoder.createInstance());
+    return new MemcacheClientBuilder<>(SerializableObjectTranscoder.getInstance());
   }
 
   /**
--- a/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
+++ b/folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
@@ -59,7 +59,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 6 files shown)


folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -59,7 +59,7 @@
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
-public class MemcacheClientBuilder<V> {
+public class MemcacheClientBuilder<V extends Serializable> {
```
spkrka: This line can and should be reverted, to avoid breaking usages that do not use Serializable classes

folsom/src/main/java/com/spotify/folsom/transcoder/SerializableObjectTranscoder.java
```diff
@@ -20,9 +20,10 @@
 import java.io.Serializable;
 import org.apache.commons.lang.SerializationUtils;
 
-public final class SerializableObjectTranscoder implements Transcoder<Serializable> {
-
-  public static final SerializableObjectTranscoder INSTANCE = new SerializableObjectTranscoder();
```
spkrka, line 25: Do not remove this fields since it is public

folsom/src/main/java/com/spotify/folsom/transcoder/SerializableObjectTranscoder.java
```diff
@@ -20,9 +20,10 @@
...
+public final class SerializableObjectTranscoder<T extends Serializable> implements Transcoder<T> {
+  public static <T extends Serializable> SerializableObjectTranscoder<T> createInstance() {
```
spkrka: rename to getInstance

folsom/src/main/java/com/spotify/folsom/transcoder/SerializableObjectTranscoder.java
```diff
@@ -20,9 +20,10 @@
...
+    return new SerializableObjectTranscoder<>();
```
spkrka: return INSTANCE

PR #196, opened 2022-03-03
docs: update readme serializable exp to use binary

liufuyang:
A serializable client will store serialized Java object as binary and it doesn't really
make sense to create a AsciiMemcacheClient and the incr and decr operation won't work either.
While the binary client's incr/decr can still be used (if get/set are not used together on those keys) as an initial value can be provided for it.


spkrka:
I do not think we should merge this.
Both Ascii and Binary protocol can be used for binary data.
Both interfaces include incr/decr which do not make sense when using a serializable transcoder

liufuyang:
Thank you. In that sense, we probably don't merge this 👍

PR #199, opened 2022-04-21
Fix maxSetLength

borislopezaraoz:
Memcached has a limit on the size of the value, which is by default 1MB. However, iiuc, this limit is not just for the value but for the whole request (key + value + flags + cas). 
As far as I can tell,  it is checked [here](https://github.com/memcached/memcached/blob/046c4bb5d8498420c13e5357c8299b60952b2595/proto_text.c#L1971-L1973), which calls [item_size_ok](https://github.com/memcached/memcached/blob/4b23988acb2151de8009d0ddc5e9d770f7bf3169/items.c#L366), which calls [item_make_header](https://github.com/memcached/memcached/blob/4b23988acb2151de8009d0ddc5e9d770f7bf3169/items.c#L165).
Folsom tries to check if the value is too large before sending the request to memcached, but [it only takes the value size into account](https://github.com/spotify/folsom/blob/master/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java#L224-L231). It should instead take the whole [request](https://github.com/spotify/folsom/blob/master/folsom/src/main/java/com/spotify/folsom/client/ascii/SetRequest.java#L99-L117) into account, to be consistent with the memcached.
Sending these requests that are bound to fail to memcached badly affects performance since they end the connection to memcached.

This PR fixes this issue by doing the correct validation (against request size not value size) before sending the request.

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
+++ b/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
@@ -223,8 +223,8 @@ public class DefaultRawMemcacheClient extends AbstractRawMemcacheClient {
   public <T> CompletionStage<T> send(final Request<T> request) {
     if (request instanceof SetRequest) {
       SetRequest setRequest = (SetRequest) request;
-      byte[] value = setRequest.getValue();
-      if (value.length > maxSetLength) {
+      int margin = 40; // see MaxSetLengthTest
+      if (setRequest.size() > maxSetLength - margin) {
         return (CompletionStage<T>)
             onExecutor(CompletableFuture.completedFuture(MemcacheStatus.VALUE_TOO_LARGE));
       }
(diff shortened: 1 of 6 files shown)


folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -223,8 +223,8 @@ private DefaultRawMemcacheClient(
   public <T> CompletionStage<T> send(final Request<T> request) {
     if (request instanceof SetRequest) {
       SetRequest setRequest = (SetRequest) request;
-      byte[] value = setRequest.getValue();
-      if (value.length > maxSetLength) {
+      int margin = 40; // see MaxSetLengthTest
+      if (setRequest.size() > maxSetLength - margin) {
```
lgylym, line 227: By any chance the extra stuff in `setRequest.size()` contribute to the `margin`?

lgylym: What I'm more interested here is that, is there some constant value that we can use, such as `value.length > maxSetLength - another_margin`, so we don't have to calculate this `size()` every time? We don't need to reach the exact limit of memcached if we can have some constant here.

borislopezaraoz:
closing this one, instead [this change](https://github.com/spotify/folsom/pull/200) will avoid closing the connection when the request is too big

PR #200, opened 2022-04-25
Handle some ASCII server errors properly

spkrka:


folsom/src/main/java/com/spotify/folsom/client/ascii/SetRequest.java
```diff
@@ -148,6 +148,12 @@ public void handle(final AsciiResponse response, final HostAndPort server) throw
       case NOT_STORED:
         succeed(MemcacheStatus.ITEM_NOT_STORED);
         return;
+      case VALUE_TOO_LARGE:
+        succeed(MemcacheStatus.VALUE_TOO_LARGE);
+        return;
+      case OUT_OF_MEMORY:
+        succeed(MemcacheStatus.OUT_OF_MEMORY);
+        return;
```
gizem969, line 156: Should these succeed or fail?
Binary client succeeds in the corresponding status, so this would be consistent with that.
But semantically aren't these failed operations? Also in the current version of the library these fail, this will be changing that behaviour.

gizem969 (approved):
I've asked one question, I have mixed feelings about it, but considering the current behaviour of the binary client, it makes sense to me.

Otherwise, LGTM 👍

folsom/src/main/java/com/spotify/folsom/client/ascii/AsciiMemcacheDecoder.java
```diff
@@ -216,7 +216,17 @@ protected void decode(final ChannelHandlerContext ctx, final ByteBuf buf, final
...
+          final String lineStr = toString(line);
+          switch (lineStr) {
+            case "SERVER_ERROR object too large for cache":
+              out.add(AsciiResponse.VALUE_TOO_LARGE);
+              return;
+            case "SERVER_ERROR out of memory storing object":
+              out.add(AsciiResponse.OUT_OF_MEMORY);
+              return;
+            default:
+              throw new IOException("Unexpected line: " + lineStr);
```
protocol7, line 228: Perhaps make the error message a bit more descriptive.

folsom/src/main/java/com/spotify/folsom/client/ascii/AsciiMemcacheDecoder.java
```diff
@@ -216,7 +216,17 @@ protected void decode(final ChannelHandlerContext ctx, final ByteBuf buf, final
           out.add(AsciiResponse.NOT_STORED);
           return;
         } else {
-          throw fail();
+          final String lineStr = toString(line);
+          switch (lineStr) {
+            case "SERVER_ERROR object too large for cache":
```
protocol7, line 221: Do we have any idea how stable these error messages are? If they change, we will change (back) the behaviour to closing the connection, which might be more unexpected than retaining it.

folsom/src/main/java/com/spotify/folsom/client/ascii/SetRequest.java
gizem969: I guess returning a completed future with the `VALUE_TOO_LARGE` status would also be consistent with [this check](https://github.com/spotify/folsom/blob/master/folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java#L227-L230) so it makes sense in that regard as well.

borislopezaraoz (approved):

folsom/src/main/java/com/spotify/folsom/client/ascii/AsciiMemcacheDecoder.java
spkrka: This is the same as it was before this change though, so I will leave it alone

folsom/src/main/java/com/spotify/folsom/client/ascii/AsciiMemcacheDecoder.java
spkrka: No idea, which is (I suppose), why we haven't had these checks before - it's not part of the memcached protocol, just something we've seen in the memcached source code:

https://github.com/memcached/memcached/blob/046c4bb5d8498420c13e5357c8299b60952b2595/proto_text.c#L1972

gizem969: Not the error message itself, but the `SERVER_ERROR` is in the protocol

https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L162-L179

Do you think it would be dangerous to catch all the `SERVER_ERROR`s and handle them the same way,  i.e. as succeeded futures with a `MemcacheStatus.SERVER_ERROR`? The protocol mentions that memcached itself closes the connection in case it's necessary, it might be good enough for us?

>In cases
  of severe server errors, which make it impossible to continue
  serving the client (this shouldn't normally happen), the server will
  close the connection after sending the error line.

spkrka: That could perhaps be done as a followup PR, I think we should try avoid to do too many things in the same change.

PR #202, opened 2022-05-26
Added support for connecting to TLS enabled memcached.

jreilly650:


folsom/src/main/java/com/spotify/folsom/MemcacheClientBuilder.java
```diff
@@ -553,6 +556,11 @@ MemcacheClientBuilder<V> withoutAuthenticationValidation() {
     return this;
   }
 
+  public MemcacheClientBuilder<V> withSSLEngineFactory(SSLEngineFactory sslEngineFactory) {
```
protocol7: Nit: make parameter final

folsom/src/main/java/com/spotify/folsom/client/tls/DefaultSSLEngineFactory.java
```diff
@@ -0,0 +1,24 @@
...
+    this(SSLContext.getDefault());
+  }
+
+  public DefaultSSLEngineFactory(SSLContext sslContext) {
+    this.sslContext = sslContext;
+  }
+
+  @Override
+  public SSLEngine createSSLEngine(String hostname, int port) {
+    SSLEngine sslEngine = sslContext.createSSLEngine(hostname, port);
```
protocol7: Not: final

folsom/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
```diff
@@ -75,6 +76,40 @@ public ReconnectingClient(
...
+        connectionTimeoutMillis,
+        charset,
+        metrics,
+        maxSetLength,
+        eventLoopGroup,
+        channelClass,
+        null);
+  }
+
+  public ReconnectingClient(
```
protocol7, line 97: Does this need to be public?

protocol7:
This looks like a great improvement, thanks for the PR! I left a few nitpick comments, but overall I think this looks reasonable. Would like @spkrka's review as well though.

Note the failed test though. Not sure if this might need anything specific in Java 8.

jreilly650:
> Note the failed test though. Not sure if this might need anything specific in Java 8.

Oh interesting.  I'll have to have a look.  I just tried it here and it is running on my java8 but not the java11 I have installed. I suspect it is related to the javax.net.ssl properties.

jreilly650:
> > Note the failed test though. Not sure if this might need anything specific in Java 8.
> 
> Oh interesting. I'll have to have a look. I just tried it here and it is running on my java8 but not the java11 I have installed. I suspect it is related to the javax.net.ssl properties.

Actually, it was working on some runs but not others.  It actually seems to be related to TLS session resumption.  Once I configure the ssl session cache on the memcached side I can no longer reproduce the issue.  Alternatively in DefaultSSLEngineFactory.createSSLEngine(String, int), if I use sslContext.createSSLEngine() instead of sslContext.createSSLEngine(hostname, port) I no longer see the issue (providing hostname and port is required for session reuse as far as I know).  I think the best thing is for me to add a `boolean allowSessionReuse` to the constructor of DefaultSSLEngineFactory to configure which `createSSLEngine` to call.  In the test, I will configure the session cache and configure for session resumption.  I'll update the PR and test a bit more tomorrow.

spkrka:
Thanks for the PR! Looks interesting, but big. I will try to take some time to look at it in the near future but I can't commit to a specific deadline.

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -125,13 +156,35 @@ public static CompletionStage<RawMemcacheClient> connect(
...
+            final SSLEngine sslEngine;
+            if (sslEngineFactory != null) {
+              sslEngine =
+                  sslEngineFactory.createSSLEngine(address.getHostText(), address.getPort());
+            } else {
+              sslEngine = null;
+            }
+
+            final ChannelHandler[] handlers;
+            if (sslEngine != null) {
```
spkrka, line 171: Here I would prefer using an arraylist and dynamically add the handlers instead of duplicating code in both paths.

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -113,6 +116,34 @@ public static CompletionStage<RawMemcacheClient> connect(
...
+        connectionTimeoutMillis,
+        charset,
+        metrics,
+        maxSetLength,
+        eventLoopGroup,
+        channelClass,
+        null);
+  }
+
+  public static CompletionStage<RawMemcacheClient> connect(
```
spkrka, line 136: Does this need to be public?

folsom/src/test/java/com/spotify/folsom/TlsIntegrationTest.java
```diff
@@ -0,0 +1,681 @@
...
+import junit.framework.AssertionFailedError;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class TlsIntegrationTest {
```
spkrka: Is this copied from the other IntegrationTest? can we unify them or have some abstract base class to reduce code duplication?

folsom/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
jreilly650: The constructor is called by MemcacheBuilderClient which is in a different package, so I think it needs to be public.

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
jreilly650: It is called from ReconnectingClient (in different package).

folsom/src/test/java/com/spotify/folsom/TlsIntegrationTest.java
jreilly650: Yes, will rework this.

jreilly650:
Would it be possible to re-run the tests on my latest change?

protocol7:
@jreilly650 kicked it off now

folsom/src/test/java/com/spotify/folsom/MemcachedServer.java
```diff
@@ -32,19 +35,63 @@ public class MemcachedServer {
       Suppliers.memoize(MemcachedServer::new)::get;
   private final String username;
   private final String password;
+  private final boolean secure;
```
spkrka, line 38: nit: rename to useTLS

spkrka (approved):
LGTM, but would also like to know what @protocol7 thinks

spkrka:
Still fails formatting:

```
Error:  Found 2 non-complying files, failing build
Error:  To fix formatting errors, run "mvn fmt:format"
Error:  Non complying file: /home/runner/work/folsom/folsom/folsom/src/test/java/com/spotify/folsom/TlsIntegrationTest.java
Error:  Non complying file: /home/runner/work/folsom/folsom/folsom/src/test/java/com/spotify/folsom/IntegrationTest.java
```

jreilly650:
> Still fails formatting:
> 
> ```
> Error:  Found 2 non-complying files, failing build
> Error:  To fix formatting errors, run "mvn fmt:format"
> Error:  Non complying file: /home/runner/work/folsom/folsom/folsom/src/test/java/com/spotify/folsom/TlsIntegrationTest.java
> Error:  Non complying file: /home/runner/work/folsom/folsom/folsom/src/test/java/com/spotify/folsom/IntegrationTest.java
> ```

Sorry, I have that fixed locally but did not push yet.  I still need to look into the intermittent test failure a little bit more.

jreilly650:
> Sorry, I have that fixed locally but did not push yet. I still need to look into the intermittent test failure a little bit more.

I spent some time looking at the intermittent failure.  So far, I have not been able to re-produce it when running the tests against memcached running locally instead of in a container, but plan to do some more test runs to see if it just takes more runs to reproduce it that way.  When running in a container, I typically see the failure about 2 runs out of 25.

I also decided to try out the openssl impl in netty to see if that made a difference, but I also see the same behavior with that.  

Based on my changes to use the openssl impl, I think I should make some small changes to the SSLEngineFactory I previously introduced.  I will change to use netty's SslContext (and calling newHandler(ByteBufAllocator) or newHandler(ByteBufAllocator, String, int) on it depending whether session reuse is desired or not).  

I will continue to look into the test failure.

paralainer:
Any plans to finish this PR? I am keen to pick it up where it's left @jreilly650 . We really need TLS support for our project.

jreilly650:
> Any plans to finish this PR? I am keen to pick it up where it's left @jreilly650 . We really need TLS support for our project.

Hi @paralainer, 
I did not get a chance to dig more into the integration test issues when using memcached in a container.  I ended up forking and working around the test issue with the plan to fix it later, but I did not get back to it yet :(

I ended up changing to use netty SslContext instead of jdk SSLContext.  I still want to try and get a PR together at some point with my changes, but I don't have time to look at it at the moment unfortunately :(

PR #204, opened 2022-09-20
Expose set request's opcode

crudmod:


folsom/src/main/java/com/spotify/folsom/client/SetRequest.java
```diff
@@ -17,4 +17,5 @@
 
 public interface SetRequest {
   byte[] getValue();
+  OpCode getOpCode();
```
spkrka: Since this is a public interface, this method needs to have a default implementation in order to be backwards compatible. Perhaps it can return NOOP by default

crudmod: makes sense, just pushed a change. 

https://github.com/spotify/folsom/pull/204/commits/2dcb63dc08b19ba1b04a10bd903b1d4a648f9f63

spkrka: Needs something like this here:
```
default OpCode getOpCode() {
 return NOOP;
}
```

crudmod: ahh cool thanks! I didn't know about default method. Just pushed a change. https://github.com/spotify/folsom/pull/204/commits/ccb6509763b8ec2e41060f752c099ddc77a71118

spkrka (approved):

PR #224, opened 2023-07-28
Add support for authenticated ASCII protocol

thiagocesarj:
This PR introduces support for the ASCII protocol authentication, as described in the [protocol.txt](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L186) definition as below:

```
Optional username/password token authentication (see -Y option). Used by
sending a fake "set" command with any key:

set <key> <flags> <exptime> <bytes>\r\n
username password\r\n

key, flags, and exptime are ignored for authentication. Bytes is the length
of the username/password payload.

- "STORED\r\n" indicates success. After this point any command should work
  normally.

- "CLIENT_ERROR [message]\r\n" will be returned if authentication fails for
  any reason.
```

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/authenticate/AsciiAuthenticator.java
+++ b/folsom/src/main/java/com/spotify/folsom/authenticate/AsciiAuthenticator.java
@@ -16,48 +16,44 @@
 
 package com.spotify.folsom.authenticate;
 
-import static com.spotify.folsom.client.Utils.unwrap;
+import static java.util.Objects.requireNonNull;
 
+import com.spotify.folsom.MemcacheAuthenticationException;
+import com.spotify.folsom.MemcacheStatus;
 import com.spotify.folsom.RawMemcacheClient;
-import com.spotify.folsom.client.ascii.GetRequest;
-import java.nio.charset.StandardCharsets;
-import java.util.concurrent.CompletionException;
+import com.spotify.folsom.client.ascii.AsciiAuthenticateRequest;
 import java.util.concurrent.CompletionStage;
 
-/**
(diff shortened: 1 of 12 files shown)


folsom/src/main/java/com/spotify/folsom/authenticate/AsciiAuthenticator.java
```diff
@@ -0,0 +1,69 @@
...
+  private final String username;
+  private final String password;
+
+  public AsciiAuthenticator(final String username, final String password) {
+    this.username = requireNonNull(username);
+    this.password = requireNonNull(password);
+  }
+
+  @Override
+  public CompletionStage<RawMemcacheClient> authenticate(RawMemcacheClient client) {
```
protocol7: Nit: make parameter final

folsom/src/main/java/com/spotify/folsom/client/ascii/AsciiAuthenticateRequest.java
```diff
@@ -0,0 +1,36 @@
...
+public class AsciiAuthenticateRequest extends SetRequest {
+
+    private static final byte[] KEY = "ASCII_AUTH".getBytes(StandardCharsets.US_ASCII);
+
+    public AsciiAuthenticateRequest(final String username, final String password) {
+        super(Operation.SET, KEY, (username + " " + password).getBytes(US_ASCII), 0, 0, 0 );
+    }
+
+    @Override
+    public void handle(AsciiResponse response, HostAndPort server) throws IOException {
```
protocol7: Nit: make parameters final

protocol7 (approved):
Looks good as far as I can tell, but will leave open for other maintainers to have a chance to review.

PR #227, opened 2023-08-16
Added support for connecting to TLS enabled memcached (new PR)

paralainer:
Picking up https://github.com/spotify/folsom/pull/202 as it seems to be abandoned, I don't have permission to push into that branch so created a new PR for that.

At this stage I merged master into the branch. I saw some intermittent test failures were reported I'll look into that as well


paralainer:
PR is ready for review. Tests are ok locally but I need approval to run them on CI

folsom/src/test/java/com/spotify/folsom/IntegrationTest.java
paralainer, line 1: Compared the diff with AbstractIntegrationTestBase.java and confirmed that all tests are up to date

paralainer:
@spkrka @protocol7 You seem to be on the previous PR for TLS support, can you please have a look at this one?
We really need TLS support and team is already considering vendoring of the library, which we obviously would like to avoid :)

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -113,6 +118,34 @@ public static CompletionStage<RawMemcacheClient> connect(
...
+        connectionTimeoutMillis,
+        charset,
+        metrics,
+        maxSetLength,
+        eventLoopGroup,
+        channelClass,
+        null);
+  }
+
+  public static CompletionStage<RawMemcacheClient> connect(
```
protocol7: Do we need the extra constructor here? This is not part of the public API so would be fine to change the existing one.

paralainer: Makes sense, fixed

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -124,14 +117,17 @@ public static CompletionStage<RawMemcacheClient> connect(
...
           @Override
-          protected void initChannel(final Channel ch) throws Exception {
-            ch.pipeline()
-                .addLast(
-                    new TcpTuningHandler(),
-                    decoder,
-
-                    // Downstream
-                    new MemcacheEncoder());
+          protected void initChannel(final Channel ch) {
```
paralainer, line 132: Did a bit of refactoring for this logic comparing to previous PR, less code

folsom/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java
```diff
@@ -75,6 +76,40 @@ public ReconnectingClient(
       final int maxSetLength,
       final EventLoopGroup eventLoopGroup,
       final Class<? extends Channel> channelClass) {
```
paralainer: @protocol7 does it make sense to remove this constructor as well?

paralainer: It was unused and not part of public API, so removed it

folsom/src/main/java/com/spotify/folsom/client/tls/DefaultSSLEngineFactory.java
```diff
@@ -0,0 +1,33 @@
...
+    this(SSLContext.getDefault(), reuseSession);
+  }
+
+  public DefaultSSLEngineFactory(final SSLContext sslContext, final boolean reuseSession) {
+    this.sslContext = sslContext;
+    this.reuseSession = reuseSession;
+  }
+
+  @Override
+  public SSLEngine createSSLEngine(String hostname, int port) {
```
protocol7: Make parameters final

paralainer: done

folsom/src/test/java/com/spotify/folsom/MemcachedServer.java
```diff
@@ -38,34 +40,64 @@ public enum AuthenticationMode {
...
       container.withExposedPorts(DEFAULT_PORT);
     }
+
+    start(authenticationMode);
+  }
+
+  private FixedHostPortGenericContainer setupContainer(
+      String username, String password, AuthenticationMode authenticationMode) {
+    final FixedHostPortGenericContainer container =
+        new FixedHostPortGenericContainer("bitnami/memcached" + ":" + MEMCACHED_VERSION);
```
protocol7:
```suggestion
        new FixedHostPortGenericContainer("bitnami/memcached:" + MEMCACHED_VERSION);
```

folsom/src/test/java/com/spotify/folsom/MemcachedServer.java
```diff
@@ -38,34 +40,64 @@ public enum AuthenticationMode {
...
+
     switch (authenticationMode) {
-      case NONE:
```
protocol7, line 67: Could this be NONE when not using TLS?

paralainer: It could, but there is no way to make this check exhaustive (fail in compile time) and for NONE we don't modify the state, so it didn't seem useful to have it. I'll return it though if you prefer to have it.

paralainer: @protocol7 any concerns here? Something else preventing this PR to be merged? 
Also can we run a CI build to make sure it's all good

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -30,28 +30,19 @@
...
-import io.netty.channel.ChannelFutureListener;
-import io.netty.channel.ChannelHandler;
-import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelInboundHandler;
-import io.netty.channel.ChannelInitializer;
-import io.netty.channel.ChannelOption;
-import io.netty.channel.ChannelPromise;
-import io.netty.channel.DefaultChannelPromise;
-import io.netty.channel.EventLoopGroup;
+import io.netty.channel.*;
```
ghost: Try to avoid wildcard imports

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
```diff
@@ -112,7 +104,8 @@ public static CompletionStage<RawMemcacheClient> connect(
       final Metrics metrics,
       final int maxSetLength,
       final EventLoopGroup eventLoopGroup,
-      final Class<? extends Channel> channelClass) {
+      final Class<? extends Channel> channelClass,
+      final SSLEngineFactory sslEngineFactory) {
```
ghost, line 120: I strongly recommend overloading here, even if the original method just calls this new one with an extra `null` in the end. This is because this change without overloading would force library users to change their code to add an extra parameter.

folsom/src/main/java/com/spotify/folsom/client/tls/DefaultSSLEngineFactory.java
```diff
@@ -0,0 +1,33 @@
+package com.spotify.folsom.client.tls;
+
+import java.security.NoSuchAlgorithmException;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+
+public class DefaultSSLEngineFactory implements SSLEngineFactory {
```
ghost, line 7: Will this interface need more than this one implementation?

folsom/src/test/java/com/spotify/folsom/MemcachedServer.java
```diff
@@ -38,34 +40,64 @@ public enum AuthenticationMode {
...
+      boolean useTLS) {
     this.username = username;
     this.password = password;
-    container = new FixedHostPortGenericContainer("bitnami/memcached:1.6.21");
+    this.useTLS = useTLS;
+
+    if (useTLS) {
+      if (authenticationMode != AuthenticationMode.NONE) {
+        throw new RuntimeException(
+            "Authentication is not currently supported with the TLS-enabled container");
```
ghost, line 78: Is this actually true? I can't test it right away, but taking a superficial look at the memcached code, TLS does not disable authentication.

paralainer: That is only about that test class, the issue is with Docker image. "memcached" image supports TLS but not authentication, "bitnami/memcached" supports authentication but not TLS, I was not able to find anything remotely popular that supports both.
I tried to create a Dockerfile based on "memcached" image but enabling SASL authentication doesn't seem trivial and steals the focus from what this PR is aiming to achive, I was going to propose another PR to improve tests after this one is merged.

folsom/src/main/java/com/spotify/folsom/client/tls/DefaultSSLEngineFactory.java
paralainer: Yes, users might want to implement their own way of constructing `SSLEngine`. The most common way is in `DefaultSSLEngineFactory` but this is not the only way.

folsom/src/main/java/com/spotify/folsom/client/DefaultRawMemcacheClient.java
paralainer: I was following @protocol7 's recommendations when I removed an extra constructor. The point is that constructor is not a part of public API, also it's highly unlikely one would construct `DefaultRawMemcacheClient` manually, builder gives a lot of extra features like retries and Ketama support

folsom-elasticache/src/main/java/com/spotify/folsom/elasticache/ElastiCacheResolver.java
```diff
@@ -181,5 +196,13 @@ public Response resolve() {
         throw new RuntimeException("ElastiCache auto-discovery failed", e);
       }
     }
+
+    private Socket createSocket() throws IOException {
+      if (useTls) {
+        return SSLSocketFactory.getDefault().createSocket();
```
paralainer, line 202: This is a bit opionated but given that the purpose of this resolver is to connect to AWS resource alternative methods won't be required anyway.
I checked that on a real application and it works perfectly fine.

paralainer:
@thiagocesarj thanks! Can you triger the build?

paralainer:
@thiagocesarj I think I found the reason for test failures - SSL data aggregation doesn't play well with memcached protocol.
Basically what happens is memcached is waiting for client to send some data to finish an operation but client doesn't send data before it has enough to fill size defined in `wrapDataSize`
Can you please run the build again, it seems to work fine for my local branch/workflow.

thiagocesarj:
@paralainer seems like we have a successful build. I will leave the PR open for a couple more days and ask a couple of people to take a second look, and if it's all good, I will prepare a release. Thanks for the contribution!

thiagocesarj (approved):

thiagocesarj:
There was an issue with a test during the master build. I will investigate, but for now, the PR has been reverted.

paralainer:
> There was an issue with a test during the master build. I will investigate, but for now, the PR has been reverted.

@thiagocesarj I'll have a look into build failure and raise another PR with a fix

PR #233, opened 2023-11-08
TLS Support

paralainer:
Another attempt to add TLS support.
Since last attempt:
* Merged with master
* Removed some redundant connection close code
* Changed approach to integration tests, useTls is now just a parameter for it.
* Ran pipeline multiple times on Github Actions, it didn't fail even once


paralainer:
@thiagocesarj hopefully this time it won't cause any issues with tests

thiagocesarj (approved):

thiagocesarj:
@paralainer thank you for the contribution! I will prepare a release for this feature as version 1.18.0.

thiagocesarj:
We had another [error](https://github.com/spotify/folsom/actions/runs/6850270319/job/18624105823#step:4:1641).

thiagocesarj:
This error doesn't make a lot of sense to me. It is essentially complaining that the handler in the pipeline did not handle the exception. But the change you've made puts SSL in the middle, and it not the last handler in the pipeline.

Also, it seems to happen only when building for master, but not when building for branch. This is a bit mysterious as well.

paralainer:
Oh, I'll have a look into the error, I'm pretty sure it's some problem with test setup, we use that code in prod for a couple of months now in a high load service with no issues

PR #246, opened 2026-02-13
expanding Connector to include address information

iralls:
This is useful for two reasons:
1. The address passed into ReconnectingClient is hardcoded and ought to be derived from where the connection is created
2. Allows us to utilize more robust client logic within Connectors we pass to ReconnectingClient. For example, a client that switches between IP addresses when connections fail.

### DIFF
--- a/folsom/src/main/java/com/spotify/folsom/reconnect/Connector.java
+++ b/folsom/src/main/java/com/spotify/folsom/reconnect/Connector.java
@@ -18,13 +18,12 @@ package com.spotify.folsom.reconnect;
 import com.spotify.folsom.RawMemcacheClient;
 import com.spotify.folsom.guava.HostAndPort;
 import java.util.concurrent.CompletionStage;
+import org.apache.commons.lang3.tuple.Pair;
 
 public interface Connector {
 
-  CompletionStage<RawMemcacheClient> connect();
-
   /**
-   * The host and port of RawMemcacheClient returned via connect(), if successful
+   * @return A memcached client along with its host and port information
    */
-  HostAndPort currentAddress();
+  CompletionStage<Pair<RawMemcacheClient, HostAndPort>> connect();
 }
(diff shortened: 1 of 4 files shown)


folsom/src/main/java/com/spotify/folsom/reconnect/Connector.java
```diff
@@ -16,8 +16,15 @@
...
 import java.util.concurrent.CompletionStage;
 
 public interface Connector {
+
   CompletionStage<RawMemcacheClient> connect();
+
+  /**
+   * The host and port of RawMemcacheClient returned via connect(), if successful
+   */
+  HostAndPort currentAddress();
```
spkrka: Adding a method to an interface may be a breaking change - it would be safer if it had a default method (could throw by default)

the comment seems wrong - it should be "the configured fixed connection address for this connector"

spkrka: Also, I think current implies it could change over time, but this should be fixed - if it's not fixed then i think we would need to make a bigger API change to return a complex object from connect, i.e. something like this?
`CompletionStage<Pair<RawMemcacheClient, HostAndPort>> connect();`

iralls: I think the return of the `Pair` would be ideal, or maybe even allowing `RawMemcacheClient` to expose the `HostAndPort`? To be clear, the only reason for having this additional field is to allow for the logging in `ReconnectingClient`, which is currently [hardcoded to a static `HostAndPort address` field](https://github.com/spotify/folsom/blob/master/folsom/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java#L162).

iralls: To follow up on this though, I kept them separate to allow for the field during failure logging [here](https://github.com/spotify/folsom/blob/master/folsom/src/main/java/com/spotify/folsom/reconnect/ReconnectingClient.java#L274), which the second approach can't cover. Do you think it's worth retaining the address in this log?

iralls: Made some changes in b79d7619e308e3b1bb1992cc7a13000ca30c7437 if you feel like taking a look

